PR #35, opened 2016-08-10
Fixed quoting of raw string literals.

svenpanne:
Although raw string literals are correctly parsed, their actual content is wrong: If I read the code in the simplecpp and cppcheck projects correctly, the intention is that the raw string literals are converted to normal string literals. To do this, one must escape special chars in the raw string literal, which is exactly what is done in this patch. Another approach would be keeping the raw string literals as they are, but I think cppcheck would need some changes then, too.

The fix needs to be pushed to cppcheck's copy, too, otherwise it gets the following code wrong:

```C++
static void bar(const char*) {}
static void baz(int) {}
int main() {
  bar(R"(")");
  baz(123);
  bar(R"(")");
}
```

Without the patch, cppcheck incorrectly complains about `baz` being never used. I guess what's going on is as follows: simplecpp doesn't quote the resulting strings correctly, so cppcheck actually sees 3 consecutive string literals, the middle one being the call to `baz`. :stuck_out_tongue:

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -315,6 +315,19 @@ bool isNameChar(unsigned char ch) {
     return std::isalnum(ch) || ch == '_' || ch == '$';
 }
 
+static std::string escapeString(const std::string &str) {
+    std::ostringstream ostr;
+    ostr << '"';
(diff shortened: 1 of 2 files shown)


danmar:
Thanks! I will copy this to cppcheck soonish.

PR #36, opened 2016-08-15
makefile: make it possible to override CXXFLAGS, add 'clean' target

matthiaskrgr:


danmar:
Thanks!

Now simplecpp.cpp will be compiled with -std=c++0x option. I want to ensure that such option is not necessary.

This -std option is ok for compiling test.cpp and main.cpp as these files will not be embedded into other applications. But I want simplecpp.cpp to be portable and easy to use.

Maybe you can refactor your commit somehow. Or maybe we can add a step in 'make test' that ensures there is no syntax errors when compiling with old -std option (-fsyntax-only).

danmar:
I don't want to apply this PR as it is so I close it.

PR #54, opened 2016-10-30
Support of #line directive

zingsheim:
Simplecpp does not support the "#line directive". This pull request fixes the problem.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -20,6 +20,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cstdlib>
 #include <limits>
 #include <list>
 #include <map>
@@ -32,6 +33,7 @@
 #include <fstream>
 #include <iostream>
 #include <stack>
+#include <string>
 
 #if defined(_WIN32) || defined(__CYGWIN__) || defined(__MINGW32__)
 #include <windows.h>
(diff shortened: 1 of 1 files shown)


simplecpp.cpp
```diff
@@ -392,8 +394,14 @@ void simplecpp::TokenList::readfile(std::istream &istr, const std::string &filen
                     loc.push(location);
                     location.fileIndex = fileIndex(cback()->str.substr(1U, cback()->str.size() - 2U));
                     location.line = 1U;
+                } else if (lastline == "# line %num%") {
+                    loc.push(location);
+                    location.line = atol(cback()->str.c_str());
```
danmar: please use the portable std::atol(). atol is the old c function that you'll get if you include stdlib.h.

simplecpp.cpp
```diff
@@ -870,7 +878,8 @@ std::string simplecpp::TokenList::lastLine(int maxsize) const {
             continue;
         if (!ret.empty())
             ret = ' ' + ret;
-        ret = (tok->str[0] == '\"' ? std::string("%str%") : tok->str) + ret;
+        ret = (tok->str[0] == '\"' ? std::string("%str%")
+               : isdigit(tok->str[0]) ? std::string("%num%") : tok->str) + ret;
```
danmar: likewise for isdigit. is it _known_ that str[0] is 0-0x7f here? or should there be a cast to unsigned char so isdigit will only get 0x00-0xff values.

zingsheim:
Done.
Tricky syntax of isdigit, etc.
Is there a cppcheck for this?

danmar:
> Tricky syntax of isdigit, etc. Is there a cppcheck for this?

Cppcheck just warns if it can see that parameter to isdigit is not 0x00-0xFF. maybe some users would like to see warnings for every std::isdigit(str[0]) when str is a std::string but I'm afraid that would generate noise when the user _knows_ that the first char is 0x00-0x7F. If we can see that the std::string comes from an unknown source, for example file input, I believe warnings would be ok.

I have thought about adding a flag in the ValueFlow::Value that says that it is network/file/user input. Then somehow the user should be able to say "treat file input as random". So if the code reads an integer from a file then Cppcheck will assume that it can have any value. If you would be interested to investigate that, feel free to do it.

When it comes to nonstandard glibc there is no check for that neither. I am afraid that portability warnings about it would be noisy - many people might not think it's good warnings even though they are correct.

PR #56, opened 2016-11-06
Refactor code to build with VS2010

amai2012:
test.cpp cannot be built as well.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -54,16 +54,16 @@ int main(int argc, char **argv) {
     simplecpp::TokenList rawtokens(f,files,filename,&outputList);
     rawtokens.removeComments();
     std::map<std::string, simplecpp::TokenList*> included = simplecpp::load(rawtokens, files, dui, &outputList);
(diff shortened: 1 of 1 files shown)


danmar:
Hmm.. I think that all C++11 is OK in main.cpp and test.cpp because these are just for simplecpp developers. So I am not sure about this change.

The simplecpp.cpp and simplecpp.h are supposed to be used in other software and I don't want to have any C++11 in these. If the user has some old compiler (VS2005, gcc3.x, etc) then I would like that simplecpp can still be imported and used in the project.

danmar:
I reverted this. the range for loops are much nicer. I hope you can use a newer compiler for simplecpp development.

danmar:
I think the README.md should be more clear.

To use simplecpp in your project, you can use an old c++98 compiler. You include simplecpp.cpp and simplecpp.h in your project and these should be c++98 compliant.

For simplecpp development a modern C++11 compiler is expected.

I recommend that simplecpp developers use cmake. Maybe we should try to remove the Makefile.

PR #70, opened 2017-05-07
Fix build with clang. Don't expose local helper function

amai2012:
subject says it all...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ testrunner:	test.o	simplecpp.o
 	$(CXX) $(LDFLAGS) simplecpp.o test.o -o testrunner
 	
 test:	testrunner	simplecpp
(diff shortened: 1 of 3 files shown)


danmar:
> Small refactoring

I don't see much improvement for now. it's more code. I would prefer to use the standard std::set constructor. I believe something like this is possible:

    const char * const altopData[] = {"and","or","bitand","bitor","not","not_eq","xor"};
    const std::set<std::string> altop(&altopData[0], &altopData[7]);

> Fix compiler options for test target

this is wrong. that build command makes sure that simplecpp.cpp is c++03 conformant. after your flag is added the test won't fail if some c++11 is added.

amai2012:
Ok, then we cannot use one set of compiler flags for test code (C++0x) and implementation...
Yes, alternate form may work, though it's doubling the data in memory.

PR #71, opened 2017-05-08
Fix build with clang + small refactoring

amai2012:
subject says it all.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -349,7 +349,7 @@ static unsigned short getAndSkipBOM(std::istream &istr)
     return 0;
 }
 
-bool isNameChar(unsigned char ch)
+static bool isNameChar(unsigned char ch)
 {
     return std::isalnum(ch) || ch == '_' || ch == '$';
 }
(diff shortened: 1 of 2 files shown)


simplecpp.cpp
```diff
@@ -1737,16 +1737,10 @@ namespace {
         }
     }
 
+    const char * const altopData[] = {"and","or","bitand","bitor","not","not_eq","xor"};
+    const std::set<std::string> altop(&altopData[0], &altopData[7]);
```
mathbunnyru: I think `const std::set<std::string> a{"ab", "cd", ...};` looks much better.

amai2012: That's C++11, isn't it? We need to stick to C++03 here.

mathbunnyru: Yes, that is c++11. Don't know why not to use c++11 or 14 though.

amai2012: Limitation is conservative. It shouldn't have higher requirements than cppcheck (https://github.com/danmar/cppcheck).

mathbunnyru: As I see, cppcheck requires c++11, so is it allowed to use c++11 features in simplecpp?
There are so many beautiful features in c++11.

danmar: Yes C++11 has better features. But this is a library. If somebody wants to use it I want it to be easy to use. If the compiler does not handle C++11 then it is not easy to use.

simplecpp.cpp
```diff
@@ -1737,16 +1739,10 @@ namespace {
         }
     }
 
+    const char * const altopData[] = {"and","or","bitand","bitor","not","not_eq","xor"};
```
mathbunnyru: I'm not sure about codestyle, but usually there is a space after a comma.

amai2012: Maybe this should have been addressed by astyle. I am having problems running it when developing on cygwin. Usually the version is not the required one, or I've seen astyle crashing on cygwin also.

danmar: I don't have a strong opinion about this one. whatever astyle outputs is proper coding style.

simplecpp.cpp
```diff
@@ -1,4 +1,4 @@
-/*
+ /*
```
danmar: extra space, typo?

PR #82, opened 2017-06-08
Sync simplecpp::simplifyPath() with cppcheck's Path::simplifyPath()

amai2012:
As suggested this PR aims to synchronize the simplifyPath() routine from cppcheck and simplecpp.

So far tests from cppcheck have been imported to simplecpp, a few have been added on top. Currently on Linux 4 are still failing, number on Windows is unknown and will be addressed later.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1723,6 +1723,7 @@ namespace simplecpp {
     }
 #endif
 
+    static const char SEP='/';
     static bool DOSDriveLetterPath(const std::string& path)
     {
         if (path.length()<2)
@@ -1733,14 +1734,14 @@ namespace simplecpp {
     {
         if (path.length()<3)
             return false;
-        return (DOSDriveLetterPath(path) && path[2]=='/');
+        return (DOSDriveLetterPath(path) && path[2]==SEP);
     }
 
-    enum PathComponentType { DOT, UPDIR, NAME };
+    enum PathComponentType { CURDIR, UPDIR, NAME };
     static PathComponentType DeterminePathComponentType(const std::string& str)
     {
         if (str==".")
-            return DOT;
+            return CURDIR;
         else if (str=="..")
             return UPDIR;
         else
@@ -1754,7 +1755,7 @@ namespace simplecpp {
         PathPartsT pathParts;
         std::string subPath;
         for (std::string::const_iterator it=path.begin(); it!=path.end(); ++it) {
-            if (*it=='/') {
+            if (*it==SEP) {
                 if (!subPath.empty()) {
                     pathParts.push_back(subPath);
                     subPath.clear();
@@ -1768,12 +1769,13 @@ namespace simplecpp {
         return pathParts;
     }
 
+    // Normalize path: resolve .., remove meaningless .
     static void NormalizePath(PathPartsT& pathParts)
     {
         PathPartsT::iterator current=pathParts.begin();
         while (current!=pathParts.end()) {
             switch (DeterminePathComponentType(*current)) {
-            case DOT:
+            case CURDIR:
                 current=pathParts.erase(current);
                 break;
             case UPDIR:
@@ -1784,7 +1786,7 @@ namespace simplecpp {
                     case UPDIR:
                         ++current;
                         break;
-                    case DOT:
+                    case CURDIR:
                     case NAME:
                         current=pathParts.erase(current-1, current+1);
                         break;
@@ -1805,14 +1807,16 @@ namespace simplecpp {
         // detect UNC path on DOS/Windows
         const bool isUnc = (0==path.compare(0,2,"\\\\")) || (0==path.compare(0,2,"//"));
         // replace backslash separators
-        std::replace(path.begin(), path.end(), '\\', '/');
+        std::replace(path.begin(), path.end(), '\\', SEP);
         // detect: DOS/Windows path including drive letter
-        const bool isAbsolute=path[0]=='/' || DOSDriveLetterPathAbsolute(path);
+        const bool isAbsolute=path[0]==SEP;
+        const bool isAbsoluteDOSriveLetterPath=DOSDriveLetterPathAbsolute(path);
         // Trailing slash indicates a directory is specified
-        const bool hasTrailingSlash=path[path.length()-1]=='/';
+        const bool hasTrailingSlash=path[path.length()-1]==SEP;
 
         // split in path components
         PathPartsT pathParts=SplitPath(path);
+        // and normalize
         NormalizePath(pathParts);
 
         // Build up the result string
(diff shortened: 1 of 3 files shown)


simplecpp.cpp
```diff
@@ -1724,11 +1724,31 @@ namespace simplecpp {
      */
     std::string simplifyPath(std::string path)
     {
+        const bool isUnc = path.compare(0,2,"//") == 0;
+
+        // Remove ./, .//, ./// etc. at the beginning
+        while (path.compare(0,2,"./") == 0) { // remove "./././"
```
danmar: this should be after `// replace backslash separators` so .\ is handled the same way

simplecpp.cpp
```diff
@@ -1724,11 +1724,31 @@ namespace simplecpp {
      */
     std::string simplifyPath(std::string path)
     {
+        const bool isUnc = path.compare(0,2,"//") == 0;
```
danmar: is `\\` unc?

danmar:
I assume this is a work-in-progress PR. It looks promising!

simplecpp.cpp
danmar: sorry.. unc is always "\\...." right? then this is the best location.

amai2012:
Yes indeed, this is work-in-progress, we shouldn't apply this yet.
And we should make sure that the expected/wanted test results are really ok, before changing the implementation...

simplecpp.cpp
```diff
@@ -1716,66 +1716,123 @@ namespace simplecpp {
         return ostr.str();
     }
 #else
-#define realFilename(f)  f
+    // Boring implementation, but could help in platform-independent tests:
```
danmar: A better comment would be something like: `for case sensitive file systems`

simplecpp.cpp
```diff
@@ -1716,66 +1716,123 @@ namespace simplecpp {
...
+    std::string realFilename(const std::string &f)
```
danmar: this function do stuff, copy the string.

simplecpp.cpp
amai2012: Well, the Windows implementation does more than this...

amai2012:
Maybe, the code is longer but better to understand imho (ok, I wrote it...)

What about the test results? Are the changes reasonable? Tests for Windows/DOS paths including drive letters are still missing.

simplecpp.cpp
danmar: As far as I know realFilename just changes the case so it matches the filesystem. Do you see any additional behaviour?

simplecpp.cpp
danmar: The macro will not do anything so it is cheaper

danmar:
I think it might be a good idea to handle currentPath also.

If currentPath is  /123/456 then paths should be simplified like:

../456 => . 
../abc  => ../abc

To allow unit tests it can for instance be a parameter.

simplecpp.cpp
amai2012: Yes, but as I wrote one might write tests only if there is a real function, cannot write tests for the macro.

danmar: it would be nice to test realFilename in windows. since there has been problems. but I am skeptic about testing it with unit tests since it needs files on the disk.

I don't say you should implement that now... but maybe a separate test script for windows could test it.

simplecpp.cpp
```diff
@@ -1716,42 +1716,127 @@ namespace simplecpp {
...
 
-        // remove "./"
-        pos = 0;
-        while ((pos = path.find("./",pos)) != std::string::npos) {
-            if (pos == 0 || path[pos - 1U] == '/')
-                path.erase(pos,2);
-            else
-                pos += 2;
+    /// split path string into components - neglecting the path separators!
+    static PathPartsT SplitPath(const std::string& path)
```
danmar: nit: first function name letter should be lowercase.

simplecpp.cpp
```diff
@@ -1716,42 +1716,127 @@ namespace simplecpp {
...
-                path.erase(0,pos+4);
-                pos = 0;
-            } else {
-                path.erase(pos1,pos-pos1+3);
-                pos = std::min((std::string::size_type)1, pos1);
+        return pathParts;
+    }
+
+    // Normalize path: resolve .., remove meaningless .
+    static void NormalizePath(PathPartsT& pathParts)
```
danmar: nit: first function name letter should be lowercase.

simplecpp.cpp
```diff
@@ -1716,42 +1716,127 @@ namespace simplecpp {
...
+    std::string simplifyPath(std::string path)
+    {
+        if (path.empty())
+            return "";
+        // detect UNC path on DOS/Windows
+        const bool isUnc = (0==path.compare(0,2,"\\\\")) || (0==path.compare(0,2,"//"));
+        // replace backslash separators
+        std::replace(path.begin(), path.end(), '\\', SEP);
+        // detect: DOS/Windows path including drive letter
+        const bool isAbsolute=path[0]==SEP;
```
danmar: I think `isAbsolute` could be `(path[0]==SEP || DOSDriveLetterPathAbsolute(path))` without changing the meaning.

simplecpp.cpp
```diff
@@ -1716,42 +1716,127 @@ namespace simplecpp {
...
+        return f;
+    }
 #endif
 
-    /**
-     * perform path simplifications for . and ..
-     */
-    std::string simplifyPath(std::string path)
+    static const char SEP='/';
+    static bool DOSDriveLetterPath(const std::string& path)
```
danmar: as far as I see this is only used in the `DOSDriveLetterPathAbsolute`.. I would integrate this.

test.cpp
```diff
@@ -1,28 +1,57 @@
...
         std::cerr << "expected:" << expected << std::endl;
         std::cerr << "actual:" << actual << std::endl;
     }
     return (expected == actual);
 }
 
-static int assertEquals(const unsigned int &expected, const unsigned int &actual, int line)
+static int succeeded_todos_counter;
+static int todos_counter;
+static int todoAssertEquals(const std::string &wanted, const std::string &expected, const std::string &actual, const std::string & function, const std::string &file, int line)
```
danmar: I would like that this is implemented in a separate PR. just add the test cases. however I don't currently see a need for TODO_ASSERT.

test.cpp
```diff
@@ -1257,6 +1286,65 @@ void simplifyPath()
     ASSERT_EQUALS("/../1.c", simplecpp::simplifyPath("/../a/../1.c"));
 }
 
+// tests transferred from cppcheck
+// https://github.com/danmar/cppcheck/blob/d3e79b71b5ec6e641ca3e516cfced623b27988af/test/testpath.cpp#L43 ff.
+void simplifyPath_cppcheck()
```
danmar: as far as I see all these test cases look good

danmar: I did not see the "sorting" of the testcases making it hard to see what corner cases are covered. or if there are duplicate tests.

do we need all tests with // ? that is 1 condition in the code that is tested with 17 assertions. if those are needed I would suggest we group all equivalent tests together. a/b a//b a///b are put together. then when reading those one does not need to re-interpret the path every time.

danmar:
I implemented most of this functionality with eff31dc75db2c7295177158e59e4e91f647a45b2. however as you say, tests with dos drive letters are still missing. I plan to look at that soon.

PR #87, opened 2017-06-24
fix some findings by cppcheck and clang

matthiaskrgr:


main.cpp
```diff
@@ -57,11 +57,11 @@ int main(int argc, char **argv)
     std::map<std::string, simplecpp::TokenList*> included = simplecpp::load(rawtokens, files, dui, &outputList);
     for (std::pair<std::string, simplecpp::TokenList *> i : included)
         i.second->removeComments();
-    simplecpp::TokenList output(files);
-    simplecpp::preprocess(output, rawtokens, files, included, dui, &outputList);
+    simplecpp::TokenList outputfiles(files);
```
danmar: hmm outputfiles is not a good name

simplecpp.cpp
```diff
@@ -1026,22 +1026,22 @@ namespace simplecpp {
          * @param output   destination tokenlist
          * @param rawtok   macro token
          * @param macros   list of macros
-         * @param files    the files
+         * @param inputfiles    the files
```
danmar: I use the name "files" everywhere for this. I'd say it's good to use the same name so it can be seen it is the same data.

amai2012:
How can we proceed here?

matthiaskrgr:
Eh, I totally forgot I already had this here, sorry! Anyway, the old mergerequest had conflicts, I'll try to fix address comments in the new one.

PR #90, opened 2017-07-31
Fix Visual Studio Warning C4458

Huemac:
- declaration of 'files' hides class member


danmar:
This will be copied to cppcheck someday. usually it happens after some bugfix.

PR #93, opened 2017-08-09
fix some cppcheck and compiler warnings

matthiaskrgr:


main.cpp
```diff
@@ -57,11 +57,11 @@ int main(int argc, char **argv)
     std::map<std::string, simplecpp::TokenList*> included = simplecpp::load(rawtokens, files, dui, &outputList);
     for (std::pair<std::string, simplecpp::TokenList *> i : included)
         i.second->removeComments();
-    simplecpp::TokenList output(files);
-    simplecpp::preprocess(output, rawtokens, files, included, dui, &outputList);
+    simplecpp::TokenList tokenList(files);
```
danmar: maybe outputTokens or something like that is better

danmar:
I don't know.. I have sympathy for Wshadow in general, it can prevent mistakes, but often we end up with less good names. maybe it's time to remove it from simplecpp/cppcheck and maybe just run it occasionally to make sure there are not mistakes.

main.cpp
matthiaskrgr: Ok, I changed it to outputTokens.

danmar:
ok well.. I'll push this.. but I am thinking that Wshadow should rather be removed.

PR #94, opened 2017-08-11
Sizeof segfault

Hinterwaeldlers:
Moved pull request from:
https://github.com/danmar/cppcheck/pull/943

A test has been included, which dies with segfault if the failure is reintroduced.


danmar:
It is not enough to not crash. If there are syntax errors then the user should be told about it with proper location info etc. Syntax errors should be reported through OutputList.

See how outputList is used in `define_invalid_1()`

danmar:
Thanks! I tweaked your code and committed it with f704c1b42ff5b83cf831e03b7170dc4e1c91ce5b. We apparently does not handle "sizeof int" well.

PR #110, opened 2017-11-20
Improve performance on windows

rsBNT:
by using the [faster ](https://ariccio.com/2015/01/13/how-fast-are-findfirstfilefindfirstfileex-and-cfilefind-actually/) FindFirstFileEx with `FindExInfoBasic` since not more is required at that spot.

Also return from `getFileName()` early if the haystack `filedata` is empty. The function is within the top 10 of CPU time consumers when running against a fairly large codebase and measuring with Visual Studio 2017 Performance Analysis.

Both improvements have led to an analysis time decrease of between 5% (solution with 14 projects, 3022 files to check) and 15% (solution with 645 projects, 21714 files to check).

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -35,7 +35,6 @@
 
 #if defined(_WIN32) || defined(__CYGWIN__) || defined(__MINGW32__)
 #include <windows.h>
-#include <VersionHelpers.h>
 #undef ERROR
 #undef TRUE
 #define SIMPLECPP_WINDOWS
@@ -1750,11 +1749,7 @@ static bool realFileName(const std::string &f, std::string *result)
     WIN32_FIND_DATAA FindFileData;
 	HANDLE hFind = INVALID_HANDLE_VALUE;
 	// only fetch once
-	static BOOL bIsWindows7OrGreater = IsWindows7OrGreater();
-	if (bIsWindows7OrGreater)
-		hFind = FindFirstFileExA(f.c_str(), FindExInfoBasic, &FindFileData, FindExSearchNameMatch, NULL, 0);
-	else
-		hFind = FindFirstFileA(f.c_str(), &FindFileData);
+	hFind = FindFirstFileExA(f.c_str(), FindExInfoBasic, &FindFileData, FindExSearchNameMatch, NULL, 0);
 
     if (INVALID_HANDLE_VALUE == hFind)
         return false;
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1747,9 +1747,7 @@ static bool realFileName(const std::string &f, std::string *result)
 
     // Lookup filename or foldername on file system
     WIN32_FIND_DATAA FindFileData;
-	HANDLE hFind = INVALID_HANDLE_VALUE;
-	// only fetch once
-	hFind = FindFirstFileExA(f.c_str(), FindExInfoBasic, &FindFileData, FindExSearchNameMatch, NULL, 0);
+	HANDLE hFind = FindFirstFileExA(f.c_str(), FindExInfoBasic, &FindFileData, FindExSearchNameMatch, NULL, 0);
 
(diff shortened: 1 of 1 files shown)


simplecpp.cpp
```diff
@@ -1752,8 +1753,15 @@ static bool realFileName(const std::string &f, std::string *result)
 
     // Lookup filename or foldername on file system
     WIN32_FIND_DATAA FindFileData;
-    HANDLE hFind = FindFirstFileA(&buf[0], &FindFileData);
-    if (hFind == INVALID_HANDLE_VALUE)
+	HANDLE hFind = INVALID_HANDLE_VALUE;
+	// only fetch once
+	static BOOL bIsWindows7OrGreater = IsWindows7OrGreater();
```
danmar: I don't want that this is static because that is not threadsafe. Is this a slow function or is it ok that it is called every time?

simplecpp.cpp
```diff
@@ -1752,8 +1753,15 @@ static bool realFileName(const std::string &f, std::string *result)
...
+	if (bIsWindows7OrGreater)
+		hFind = FindFirstFileExA(&buf[0], FindExInfoBasic, &FindFileData, FindExSearchNameMatch, NULL, 0);
```
danmar: I don't understand this code. Why do you only want to call this function in newer windows? Did it exist previously but did not work "properly"?

rsBNT: This was because previously i also used FIND_FIRST_EX_LARGE_FETCH which is only supported by >Win7 systems

simplecpp.cpp
```diff
@@ -1746,20 +1746,15 @@ static bool realFileName(const std::string &f, std::string *result)
     if (!alpha)
         return false;
 
-    // Convert char path to CHAR path
-    std::vector<CHAR> buf(f.size()+1U, 0);
```
danmar: must look a little at this. I thought CHAR could be 16-bit also.

I want to handle non-plain-ASCII filenames also such as åäö.cpp etc. But I don't really remember how well tested and well implemented this is.

simplecpp.cpp
rsBNT: I will benchmark FIND_FIRST_EX_LARGE_FETCH again and if its unnecessary, remove the whole check alltogether

simplecpp.cpp
amai2012: https://msdn.microsoft.com/de-de/en-en/library/windows/desktop/dd374131(v=vs.85).aspx indicates that CHAR is 8-bit always.

simplecpp.cpp
amai2012: Move static to file scope to resolve the multi-threading issue!?

rsBNT:
Im not sure how/where i saw the performance increase, i cant seem to reproduce it with small projects. I assume, i didnt properly clear the caches or something.

I did remove the OS check because [FIND_FIRST_EX_LARGE_FETCH](https://blogs.msdn.microsoft.com/oldnewthing/20131024-00/?p=2843) didnt show any improvements, but again, i cant find a good test scenario to measure the performance.

rsBNT:
Ok, did some more measurements and this now makes sense:

**main** (no cache clearing)
> 1. run: 4m21.696s 4m34.327s 4m31.609s -> avg: 4m29.121s
> 2. run: 0m11.487s 0m5.422s  0m5.423s  -> avg: 0m7.444s

**improvePerfOnWin** (no cache clearing)
> 1. run: 4m15.029s 4m16.689s 4m16.656s -> avg: 4m16.125s
> 2. run: 0m10.658s 0m5.376s  0m5.375s  -> avg: 0m7.136s

So this code is actually faster when the files have been processed by the filesystem, which is not unlikely. Since there is no penalty in the uncached case (if, then small perf gain there as well) id still propose to merge this and see if anybody with big projects notices it.

PR #117, opened 2018-02-03
Fixed potential null pointer dereference.

orbitcowboy:
nextSkipComment() can return a NULL pointer.


danmar:
I think the original code is safe. If either parameter is null then `sameline()` returns false.

PR #132, opened 2018-05-27
Test case trying to analyze https://trac.cppcheck.net/ticket/8585 - a…

amai2012:
…n issue about thwrowing exceptions of type simplecpp::Output. 
!!DO NOT APPLY/MERGE!!


amai2012:
It was for demonstration purposes only.

PR #140, opened 2018-12-07
Fix token pasting in macro when argument contains spaces. Fixes #124

### LINKED ISSUE #124: Wrong token pasting/concatenation when token contains spaces
Stumbled over this issue when i added a define for the Qt macro Q_D(), see Cppcheck ticket https://trac.cppcheck.net/ticket/8479.
For this C code:
```
#define CONCAT(tok) tok##suffix

CONCAT(Test);
CONCAT(const Test);
```
gcc outputs:
```
$ gcc -E token_pasting_with_space.c
# 1 "token_pasting_with_space.c"
# 1 "<built-in>"
# 1 "<command-line>"
# 1 "token_pasting_with_space.c"


Testsuffix;
const Testsuffix;
```
simplecpp outputs:
```
$ ./simplecpp token_pasting_with_space.c


Testsuffix ;
constTestsuffix ;
```
The space between "const" and "Test" is removed.
Not sure what the standard says about such a case but since Qt and gcc are widely used i guess it should work like Qt expects it and gcc handles it to not break things.

versat:


danmar:
Should it be 1 token? Probably there should be 1 token `const` and 1 token `Testsuffix`.

versat:
Hmm, would be somewhat strange if it is 1 token. So do you mean that calling `expandArgStr` is already wrong in this case? I must admit that i do not fully understand the code.

danmar:
I don't remember the details.. but it seems a bit strange to call expandArgStr in this case.

versat:
Would it be ok to add a TODO or FIXME comment to the code i added or would you prefer to directly fix the issue in another way?

danmar:
could you investigate how gcc outputs this? As 2 tokens or as 1.

Does it generate/reference a `Testsuffix` symbol?

danmar:
> Looks like gcc outputs two tokens const and Testsuffix.

Yes I agree!

versat:
I close this PR because i do not really have an idea how to fix it in a better way.

PR #143, opened 2019-01-25
Cache results of FindFirstFileExA to improve performance on Windows

mitsujin:
This adds a cache to the realFileName call to improve performance of check times on Windows. 
Prior to this change, cppcheck was taking 30 minutes to run on a codebase consisting of 87 solutions, 594 projects and 18000 files (using -j8). 
After this change, cppcheck now takes 12.7 minutes (-j8). 

We have been using cppcheck 1.85 with this change since October last year without issues in our nightly. 

image

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1880,7 +1880,7 @@ public:
         CloseHandle(m_mutex);
     }
 
-    bool getRealPathFromCache(std::string const& path, std::string* returnPath)
+    bool getRealPathFromCache(const std::string& path, std::string* returnPath)
     {
         ScopedLock lock(m_mutex);
 
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1840,36 +1840,73 @@ namespace simplecpp {
 
 #ifdef SIMPLECPP_WINDOWS
 
+class ScopedLock
+{
+public:
+    explicit ScopedLock(HANDLE mutex)
+        : m_mutex(mutex)
+    {
+        if (WaitForSingleObject(m_mutex, INFINITE) == WAIT_FAILED)
+            throw std::runtime_error("cannot lock the mutex");
+    }
+
+    ~ScopedLock()
+    {
+        ReleaseMutex(m_mutex);
+    }
+
+private:
+    ScopedLock& operator=(const ScopedLock&);
+    ScopedLock(const ScopedLock&);
+
+    HANDLE m_mutex;
+};
+
 class RealFileNameMap
 {
 public:
-    bool getRealPathFromCache(std::string const& path, std::string& returnPath)
+    RealFileNameMap()
     {
-        std::unique_lock<std::mutex> lock(mMutex);
-        
-        auto it = mFileMap.find(path);
-        if (it != mFileMap.end())
+        m_mutex = CreateMutex(NULL, FALSE, NULL);
+
+        if (!m_mutex)
         {
-            returnPath = it->second;
+            throw std::runtime_error("cannot create the mutex handle");
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


simplecpp.cpp
```diff
@@ -1839,8 +1839,37 @@ namespace simplecpp {
 }
 
 #ifdef SIMPLECPP_WINDOWS
+
+class RealFileNameMap
+{
+public:
+    bool getRealPathFromCache(std::string const& path, std::string& returnPath)
+    {
+        std::unique_lock<std::mutex> lock(mMutex);
```
danmar: In this project we still use C++03 to have maximum portability. And `std::unique_lock` is a C++11 feature.

I would prefer to use the windows function to lock instead of `std::unique_lock`. Can you please try this and if it does not work well.. let me know.

I guess you can use the functions described here:
https://docs.microsoft.com/en-us/windows/desktop/sync/using-mutex-objects

simplecpp.cpp
```diff
@@ -1839,8 +1839,37 @@ namespace simplecpp {
...
+        mFileMap[path] = actualPath;
+    }
+    
+    std::map<std::string, std::string> mFileMap;
+    std::mutex mMutex;
+};
+
 static bool realFileName(const std::string &f, std::string *result)
 {
+    static RealFileNameMap realFileNameMap;
```
danmar: This is not thread safe, as far as I know. I am pretty sure that the `realFileNameMap` will be initialized when this function is executed the first time (without locking mechanism). It should be safer if the variable is put in global scope.

simplecpp.cpp
```diff
@@ -1852,19 +1881,23 @@ static bool realFileName(const std::string &f, std::string *result)
             break;
         }
     }
-
+
```
danmar: you can remove these spaces

simplecpp.cpp
```diff
@@ -1839,8 +1839,37 @@ namespace simplecpp {
 }
 
 #ifdef SIMPLECPP_WINDOWS
+
+class RealFileNameMap
+{
+public:
+    bool getRealPathFromCache(std::string const& path, std::string& returnPath)
```
danmar: Coding style; I prefer that output parameters are pointers. That convention is used in many places in Cppcheck/simplecpp but maybe not all.

simplecpp.cpp
```diff
@@ -1839,8 +1839,37 @@ namespace simplecpp {
...
+        auto it = mFileMap.find(path);
+        if (it != mFileMap.end())
+        {
+            returnPath = it->second;
+            return true;
+        }
+        return false;
+    }
+    
+    void addToCache(std::string const& path, std::string const& actualPath)
```
danmar: for consistency, please put the `const` before the type.

simplecpp.cpp
```diff
@@ -1839,8 +1839,37 @@ namespace simplecpp {
...
 #ifdef SIMPLECPP_WINDOWS
+
+class RealFileNameMap
+{
+public:
+    bool getRealPathFromCache(std::string const& path, std::string& returnPath)
+    {
+        std::unique_lock<std::mutex> lock(mMutex);
+        
+        auto it = mFileMap.find(path);
```
danmar: `auto` is also not allowed.

danmar:
Also; :+1 for making this PR in the `simplecpp` repo.

mitsujin:
Thanks for the review! I'll make the changes tonight. We have quite a bit of legacy code to support which does have a lot of include references unfortunately.

mitsujin:
@danmar Please merge this in if you are happy with the changes.

danmar:
Thanks!

PR #144, opened 2019-02-19
more caching for file handling

HolgiHo:
This adds further caching to file name handling and file open/exists calls.

Situation: I am running Cppcheck on a large solution with ~5000 cpp files that uses lots of vendor libraries (>10000 header files) residing in more than >20 different include paths (-I).
My caching extensions decrease checking time from 42 mins (-j24) to 4 minutes (-j24), so performance gain is more than 10x.
The key performance problem was that each #include path spec must be checked in combination with each compiler include path (-I) and simplecpp is trying to open the same (mostly non-existing) files over and over again.

Initial caching has recently been added by [mitsujin](/mitsujin) #143 to realFileName(const std::string &f, std::string *result)

Additional caching is now done in:
- realFilename(const std::string &f)  (outer function)
- simplecpp::simplifyPath
- openHeader(...) caches whether file exists (This is only beneficial together with caching in simplifyPath)

All caches must be thread-safe. [mitsujin](/mitsujin) used a Win32 mutex for this purpose, which is extremely slow due to its cross-process nature. Cppcheck could not use more than 10 cores effectively. I replaced it by a critical section object, which is process local. Cppcheck now uses all my 24 cores.

As simplecpp still requires c++0x compatibility, I could not use cross-platform C++11 multi-threading and locking features. Therefore I disabled caching completely for non-Windows platforms (no negative performance effects expected).

To later provide caching for all platforms, either C++11 features should be enabled or each platform must be handled separately.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1843,67 +1843,91 @@ namespace simplecpp {
 class ScopedLock
 {
 public:
-    explicit ScopedLock(HANDLE mutex)
-        : m_mutex(mutex)
+    explicit ScopedLock(CRITICAL_SECTION& criticalSection)
+        : m_criticalSection(criticalSection)
     {
-        if (WaitForSingleObject(m_mutex, INFINITE) == WAIT_FAILED)
-            throw std::runtime_error("cannot lock the mutex");
+        EnterCriticalSection(&m_criticalSection);
     }
 
     ~ScopedLock()
     {
-        ReleaseMutex(m_mutex);
+        LeaveCriticalSection(&m_criticalSection);
     }
 
 private:
     ScopedLock& operator=(const ScopedLock&);
     ScopedLock(const ScopedLock&);
 
-    HANDLE m_mutex;
+    CRITICAL_SECTION& m_criticalSection;
 };
 
-class RealFileNameMap
+template <typename TKey, typename TValue> class ThreadSafeCache
 {
 public:
-    RealFileNameMap()
+    typedef std::map<TKey, TValue> TMap;
+    
+    ThreadSafeCache()
     {
-        m_mutex = CreateMutex(NULL, FALSE, NULL);
-
-        if (!m_mutex)
-        {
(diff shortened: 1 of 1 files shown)


simplecpp.cpp
```diff
@@ -1843,67 +1843,91 @@ namespace simplecpp {
...
 private:
     ScopedLock& operator=(const ScopedLock&);
     ScopedLock(const ScopedLock&);
 
-    HANDLE m_mutex;
+    CRITICAL_SECTION& m_criticalSection;
 };
 
-class RealFileNameMap
+template <typename TKey, typename TValue> class ThreadSafeCache
```
danmar: I am not sure, you generalize the code but it seems the original class name and method names were more explicit... yes this could be useful in the future for some other purpose. But right now it is a file name map, the key type is a string and the value type is string. It is overkill with a template here.

Your changes are not wrong but I suggest that the original class name and method names are restored.

simplecpp.cpp
```diff
@@ -1938,11 +1962,15 @@ static bool realFileName(const std::string &f, std::string *result)
     return true;
 }
 
+static ThreadSafeCache<std::string, std::string> realPathMap;
```
danmar: It seems strange to me that we want to have different maps for different `realFileName` functions. Do you see wrong behavior when reusing the same map?

danmar:
hmm.. one idea I have is that you could just replace Mutex with critical section in one first PR. That seems ok. And then we could continue with the other updates..

I let you decide. If you want to keep it all in one PR the feel free to do that.

HolgiHo:
Ok, I agree to split up the PR. I'm going to prepare the critical section part tomorrow.
Then comes the realFilename part. You are right, we should verify any implications carefully here.
By the way, I introduced the template to also store a map (string -> bool) for the fileExistsCache.

HolgiHo:
PR closed (replaced by #145, #146 and later PRs)

PR #152, opened 2019-02-28
Cache non-existing header files to skip expensive file open call

### LINKED ISSUE #151: Bad performance when using many -I directives
@danmar Thanks for merging #145 and #150 into master. On my large solution (~5000 cpp files, using >10000 header files from external libraries) this already gives me a speedup of factor 3.

However, there is more potential for performance improvement (again factor ~3):
- Every time when resolving an #include path, cppcheck checks if a such a file exists at all locations given by compiler include paths (-I).
- I am using about 25 such -I directives due to the large number of external libraries.
- That means, simplecpp must check up to 25 locations for each #include directive, but usually the required header file exists only in 1 of these locations.
- Profiling in Visual Studio showed that ~40% of the total Cppcheck runtime is spent in ifstream.open

I could reach a speedup of factor ~3 by implementing the following:
- in openHeader: remember all file paths of files known not to exist in a "nonExistingFilesSet"
- in simplecpp::simplifyPath: cache results in a map (path -> simplifiedPath)

I already submitted the necessary changes in PR #144, but we decided to split up this PR into smaller pieces.

A word about thread-safety:
- In contrast to function realFilename, which is only used on Windows, the 2 new caches are used on all platforms (openHeader and simplifyPath are called on all platforms).
- At first sight this would mean we would have to implement a platform-independent synchronization mechanism for cache access. This would be difficult without C++11 std lib.
- However, we are lucky: Cppcheck uses multi-threading only on Windows platform. On Linux/Unix it uses forking, i.e. we simply can omit synchronization on Linux/Unix, because each cache is local to one forked process. (Unfortunately, this means caching is less effective.)

@danmar If you agree, I would submit another PR implementing the described cache features. I simply wanted to hear your opion before :-) Thank you!

HolgiHo:
This PR fixes #151: Caches the file paths of all non-existing header files in a thread-safe set.

As agreed upon in #151, caching is enabled for Windows only.

Time measurements showed that caching the results of simplecpp::simplifyPath does not decrease checking time significantly, so I did not change anything in simplifyPath.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2164,23 +2164,74 @@ static const simplecpp::Token *gotoNextLine(const simplecpp::Token *tok)
     return tok;
 }
 
+#ifdef SIMPLECPP_WINDOWS
+
+class NonExistingFilesCache {
+public:
+    NonExistingFilesCache() {
+        InitializeCriticalSection(&m_criticalSection);
+    }
+
+    ~NonExistingFilesCache() {
+        DeleteCriticalSection(&m_criticalSection);
+    }
+
+    bool contains(const std::string& path) {
+        ScopedLock lock(m_criticalSection);
+        return (m_pathSet.find(path) != m_pathSet.end());
+    }
+
+    void add(const std::string& path) {
+        ScopedLock lock(m_criticalSection);
+        m_pathSet.insert(path);
+    }
+
+private:
+    std::set<std::string> m_pathSet;
+    CRITICAL_SECTION m_criticalSection;
+};
+
+static NonExistingFilesCache nonExistingFilesCache;
+
+#endif
+
+static std::string _openHeader(std::ifstream &f, const std::string &path)
+{
+#ifdef SIMPLECPP_WINDOWS
+    std::string simplePath = simplecpp::simplifyPath(path);
+    if (nonExistingFilesCache.contains(simplePath))
+        return "";  // file is known not to exist, skip expensive file open call
+
+    f.open(simplePath.c_str());
+    if (f.is_open())
+        return simplePath;
+    else
+    {
+        nonExistingFilesCache.add(simplePath);
+        return "";
+    }
+#else
+    f.open(path.c_str());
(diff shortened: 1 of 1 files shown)


danmar:
I wonder if the cache changes might have caused problems;
http://trac.cppcheck.net/ticket/9003

Do you have any opinion?

HolgiHo:
Assuming that file system contents remain constant during the whole Cppcheck run, function realFilename operates without any side effects (like in functional programming). This is definitely true for absolute file paths, but also for relative file paths, because the Cppcheck working directory (as base directory for resolving relative paths against) does not change at any time.

Thus realFilename qualifies for the caching mechanism implemented.

Nevertheless we should await user response for the Trac ticket.

HolgiHo:
@danmar After analyzing the realFilename code in depth, I found 3 issues:
1. The name of "getRealPathFromCache" is misleading. In case of inner function realFileName(const std::string &f, std::string *result), the cache stores the last file name only (example entry "a/B" -> "b"). Thus we cannot use the same cache for the outer function realFilename(const std::string &f), which stores whole path results (example: "a/B" -> "a/b"). I cannot say if this is causing http://trac.cppcheck.net/ticket/9003  without more information from the ticket issuer. Nevertheless we should correct this.
1. An older bug: The loop condition in the first for-statement in inner function realFileName(const std::string &f, std::string *result) should be: pos <= f.size(); and not "<". Otherwise we never check whether f[0] is a letter.
1. When on cygwin, getting the realFilename for an absolute path like /cygdrive/c/... will not work as we are bypassing cygwin file path transformation logic by calling FindFirstFileExA directly. When working with relative paths to the solution or cpp file, this works (tested).

Do you have any suggestions? I would prefer fixing no. 1 and 2 first and 3 maybe later. It would be cool to also get this PR merged soon as I (and others) are eager to improve cppcheck performance on large solutions.

Thank you for all the effort with developing cppcheck!

danmar:
I applied this now. Sounds good to fix the bugs. It was good catches!

PR #155, opened 2019-03-09
Merge string and character literal with prefix

rikardfalkeborn:
Previously, simplecpp would output the character `L'a'` as two tokens `"L"`
and `"'a'"`. A user of the tokens has no way of knowing if they were
originally `"L'a'"` och `"L 'a'"`.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -327,6 +327,15 @@ static void ungetChar(std::istream &istr, unsigned int bom)
         istr.unget();
 }
 
+static unsigned char prevChar(std::istream &istr, unsigned int bom)
+{
+    ungetChar(istr, bom);
+    ungetChar(istr, bom);
+    unsigned char c = readChar(istr, bom);
+    readChar(istr, bom);
+    return c;
+}
+
 static unsigned short getAndSkipBOM(std::istream &istr)
 {
     const int ch1 = istr.peek();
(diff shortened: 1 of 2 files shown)


simplecpp.cpp
```diff
@@ -545,31 +555,34 @@ void simplecpp::TokenList::readfile(std::istream &istr, const std::string &filen
...
-                    back()->setstr(cback()->str().substr(0, cback()->str().size() - 1));
-                    push_back(new Token(currentToken, location)); // push string without newlines
-                }
+                if (hasR)
+                    currentToken = escapeString(currentToken);
+                currentToken.insert(0, prefix);
+                back()->setstr(currentToken);
                 location.adjust(currentToken);
                 if (currentToken.find_first_of("\r\n") == std::string::npos)
                     location.col += 2 + 2 * delim.size();
```
rikardfalkeborn: Hm, this location might no longer be correct, I'll take a look.

rikardfalkeborn: Also,
```C++
#define A L"a\
b"
A;
```
is broken.

danmar: ok thanks for looking into this...

simplecpp.cpp
```diff
@@ -327,6 +327,15 @@ static void ungetChar(std::istream &istr, unsigned int bom)
         istr.unget();
 }
 
+static unsigned char prevChar(std::istream &istr, unsigned int bom)
+{
+    ungetChar(istr, bom);
+    ungetChar(istr, bom);
```
HolgiHo: This line introduces the following problem at least on Windows/MSVC 2017:
If a string constant starts exactly at the _INTERNAL_BUFSIZ = 4096 bytes (constant in ucrt\inc\corecrt_internal_lowio.h), simplecpp outputs
syntax error: No pair for character ("). Can't process file. File is either invalid or unicode, which is currently not supported.

Analysis: The problem occurs in new method prevChar that tries to unget 2 characters. When trying to unget the second character, the internal buffer is empty, so the call fails and sets badbit. The following call to readChar then also fails and sets failbit.

Example input file (GitHub does not support .cpp files, so please rename .txt to .cpp)
[StringConstant.txt](https://github.com/danmar/simplecpp/files/3017659/StringConstant.txt)

rikardfalkeborn: Yeah, that function did not give my a warm and fuzzy feeling when I implemented it...

Can you test #160? I was not able to reproduce this on Linux. Also, this means that reading of utf-16 files is probably broken as well, since there is no guarantee that unget can be called twice in a row?

PR #160, opened 2019-03-28
Fix check if previous read character was space

rikardfalkeborn:
The prevChar was broken in the sense that there is no guarantee that
unget can be called more than once on a stream. This lead to problems in
some cases on some operative systems. Fix this by saving the state
before reading the character. Note that prevIsSpace is not valid in the
entire while loop, if characters are read on other places than at the
start of the while loop, the value will be wrong. This is the case in
the only place that it is being used.

This feels like a very brittle solution...

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -327,15 +327,6 @@ static void ungetChar(std::istream &istr, unsigned int bom)
         istr.unget();
 }
 
-static unsigned char prevChar(std::istream &istr, unsigned int bom)
-{
-    ungetChar(istr, bom);
-    ungetChar(istr, bom);
-    unsigned char c = readChar(istr, bom);
-    readChar(istr, bom);
-    return c;
-}
-
 static unsigned short getAndSkipBOM(std::istream &istr)
 {
     const int ch1 = istr.peek();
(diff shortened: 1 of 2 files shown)


HolgiHo:
This indeed solves "unget twice fails" issue described in #155.

There is a double semicolon in
`unsigned char ch = 'a';;`

As you stated, having a variable prevIsSpace that is not always valid, is not really beautiful.

I think a more comprehensive solution would require to change the methods accessing the istream (readChar, ungetChar, but not peekChar) to store some "previous character read" state.

rikardfalkeborn:
Another solution would be to not first create the prefix tokens (`u8` etc) and then join them with the string would be to create everything in one go. That would remove the need for checking if the previous char was a space or not.

danmar:
> Another solution would be to not first create the prefix tokens (u8 etc) and then join them with the string would be to create everything in one go. That would remove the need for checking if the previous char was a space or not.

Did you investigate using the `Token::column` property? If you create the tokens `u8` and `"abc"` and then see that the columns diff is 2.. then there is no space between the tokens. For normal code this would be the approach I'd use... I just don't remember exactly how this works in macros..

rikardfalkeborn:
> > Another solution would be to not first create the prefix tokens (u8 etc) and then join them with the string would be to create everything in one go. That would remove the need for checking if the previous char was a space or not.
> 
> Did you investigate using the `Token::column` property? If you create the tokens `u8` and `"abc"` and then see that the columns diff is 2.. then there is no space between the tokens. For normal code this would be the approach I'd use... I just don't remember exactly how this works in macros..

Thanks, that seems to work like a charm, even for macros. Tests `multiline8` and `multiline9` even seems to test this.

danmar:
After merging this I thought about a pitfall. You do not check that they are on the same line:

    u8
      "abc"

rikardfalkeborn:
Right. I'll send another PR.

rikardfalkeborn:
> After merging this I thought about a pitfall. You do not check that they are on the same line:
> 
> ```
> u8
>   "abc"
> ```

Created #161.

PR #163, opened 2019-04-03
several optimizations

firewave:


danmar:
do you have some real life measurements?

firewave:
Running with -j1 to see the actual impact - the library I used to profile and some random ones.

openni-sensor-primesense
```
real    0m4.441s
user    0m3.859s
sys     0m0.516s
```
```
real    0m3.844s
user    0m3.297s
sys     0m0.531s
```

s390-tools
```
real    0m5.095s
user    0m4.594s
sys     0m0.500s
```
```
real    0m4.660s
user    0m3.938s
sys     0m0.703s
```

gabedit
```
real    0m41.950s
user    0m39.781s
sys     0m2.172s
```
```
real    0m38.831s
user    0m36.672s
sys     0m2.125s
```

firewave:
Okay, now with just the optimization ins this PR. It does not effect all scans.

openni-sensor-primesense 4.7 5.1
gabedit 44.9 44.6
s390-tools 5.5 5.5

Strangely gabedit gives about 2000 warnings less with the changes, so I need to look into this.

simplecpp.cpp
```diff
@@ -524,7 +524,8 @@ void simplecpp::TokenList::readfile(std::istream &istr, const std::string &filen
             ch = readChar(istr,bom);
             while (istr.good()) {
                 currentToken += ch;
-                if (currentToken.size() >= 4U && endsWith(currentToken, "*/"))
+                static const std::string commentEnd("*/");
```
danmar: as far as I remember we try to make simplecpp threadsafe. So a static local variable should be avoided.

simplecpp.cpp
```diff
@@ -131,7 +131,7 @@ const std::string simplecpp::Location::emptyFileName;
 
 void simplecpp::Location::adjust(const std::string &str)
 {
-    if (str.find_first_of("\r\n") == std::string::npos) {
+    if (!strpbrk(str.c_str(), "\r\n")) {
```
danmar: I find it strange that this would be an optimisation. Is strpbrk optimised better?

However I think it obfuscates the code. I do not intuitively know how strpbrk works.

firewave: callgrind showed and there's this
https://www.codeproject.com/Articles/18473/find-first-of-A-performance-pitfall-among-the-STL#A30. I used the latest GCC so it seems this is still true although that article is 12 years old.

`strpbrk` is the same as `find_first_of`.

simplecpp.cpp
firewave: Good point. Would a global one be okay?

danmar: yes

firewave:
I tried some other measurements and those were a bit inconclusive. I will look into this in the next days.

simplecpp.cpp
firewave: Running `simplecpp` on `-Ilib lib/valueflow.cpp`:
GCC `335,129,681` -> `325,111,787`
Clang `323,433,151` -> `313,133,608`

firewave:
Running `simplecpp` with `-Ilib -D __GNUC__ lib/valueflow.cpp`:
GCC `335,129,681` -> `318,581,229`
Clang `323,433,151` -> `312,465,341`

firewave:
Using `lib/valueflow.cpp` from 0a99e3b6fc94b86991b38519e77a9ad690e76f80 with `-q -Ilib -D __GNUC__ lib/valueflow.cpp`.

baseline
GCC 11 `337,505,121`
Clang 13 `325,815,035`

Location::adjust()
GCC 11 `327,385,765`
Clang 13 `315,411,381`

endsWith()
GCC 11 `320,802,714`
Clang 13 `314,737,794`

lastLine()
GCC 11 `320,639,877`
Clang 13 `314,696,243`

unordered_map
GCC 11 `306,006,031`
Clang 13 `299,186,851`

flags memchr
GCC 11 `302,776,749`
Clang 13 `296,127,177`

flags copy ctor
GCC 11 `297,134,127`
Clang 13 `291,144,265`

FYI about a third of the instructions is made up of the new/delete of the `Token` objects and the `std::istream` reading.

PR #164, opened 2019-05-01
added move constructor and assignment for TokenList

firewave:
gets rid of copy of result in Preprocessor::preprocess()

reduces Ir from 14.98% to 12.70% with package pion according to callgrind


firewave:
Another real work example:

https://github.com/mamedev/mame/blob/master/3rdparty/bgfx/3rdparty/glslang/glslang/MachineIndependent/Constant.cpp

before:
real    0m5.568s
user    0m5.453s
sys     0m0.078s

after:
real    0m4.830s
user    0m4.734s
sys     0m0.063s

PR #166, opened 2019-05-04
 updated CMake to include warnings and syntax check

firewave:
also removed -Wabi since it will no longer warn by itself will report the following for each file:
```
cc1plus: warning: -Wabi won't warn about anything [-Wabi]
cc1plus: note: -Wabi warns about differences from the most up-to-date ABI, which is also used by default
cc1plus: note: use e.g. -Wabi=11 to warn about changes from GCC 7
```

add -Wundef since it is a useful warning

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,11 @@
 cmake_minimum_required (VERSION 3.5)
 project (simplecpp LANGUAGES CXX)
 
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
     add_compile_options(-Wall -Wextra -pedantic -Wcast-qual -Wfloat-equal -Wmissing-declarations -Wmissing-format-attribute -Wredundant-decls -Wshadow -Wundef)
 endif()
 
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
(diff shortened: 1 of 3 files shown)


firewave:
The `Makefile` uses `-std=c++03` for the syntax check and `.travis.yml` uses `-std=c++98` - which is the intended one?

danmar:
> The Makefile uses -std=c++03 for the syntax check and .travis.yml uses -std=c++98 - which is the intended one?

I don't know the difference.. but I guess c++03 is fine.

firewave:
See https://en.wikipedia.org/wiki/C%2B%2B03

The problem though is, that `__cplusplus` is `199711L` for `-std=c++98` and `-std=c++03`. So code-wise you cannot differentiate them at all.

Also when using `CXX_STANDARD` it will not accept `03` as a valid value - see https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html

So C++03 is not that _portable_ after all...

danmar:
> So C++03 is not that portable after all.

I do not have a strong opinion. If you think c++98 would be more portable then please feel free to adjust.

The switch to C++11 will probably come some day. Hard to say when. But right now I don't know if it hurts to keep it c++98/c++03 compatible.

firewave:
I am all for keeping it compatible for the time being, but I just don't know which syntax. As there's no way to properly differentiate between C++98 and C++03 I would go with the first one. In that case there would still be a few warnings to fix.

And you just merged it ... :)

PR #172, opened 2019-07-24
Added octal numbers recognition

o-bolshakov:
Working with cppcheck I've faced off the issue in case using macros with octal numbers (access modes for files in UNIX). So I've create patch for simplecpp.cpp to fix this issue. The patch has been tested for a while in my own cppcheck build. I guess that it looks stable to be included in master.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -45,6 +45,11 @@ static bool isHex(const std::string &s)
     return s.size()>2 && (s.compare(0,2,"0x")==0 || s.compare(0,2,"0X")==0);
 }
 
+static bool isOct(const std::string &s)
+{
+    return s.size()>1 && (s.compare(0,1,"0")==0);
+}
+
 
 static const simplecpp::TokenString DEFINE("define");
 static const simplecpp::TokenString UNDEF("undef");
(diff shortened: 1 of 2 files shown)


simplecpp.cpp
```diff
@@ -45,6 +45,11 @@ static bool isHex(const std::string &s)
     return s.size()>2 && (s.compare(0,2,"0x")==0 || s.compare(0,2,"0X")==0);
 }
 
+static bool isOct(const std::string &s)
+{
+    return s.size()>1 && (s.compare(0,1,"0")==0);
```
danmar: please make this function more strict.. for instance, it should return false for `0xff`, `0.5`.. etc

danmar: If you check that second char is a digit that is fine for me.

o-bolshakov: Hi Daniel, you're right. I'll add test and check for second char. Stay tuned.

o-bolshakov: Is this will be Ok to do as additional commit? Or should it be better if I create another branch with all of this in one commit and when do another pull request?

danmar: please commit the fixes in your `octal-numbers` branch. When you push it, then it should be shown here in this Pull Request. You may need to use `--set-upstream` but git will tell you about it then.

o-bolshakov: Already done.

o-bolshakov: I've also checked new constFold() tests without isOct() and they fault. So it seems that isOct() works as expected.

PR #180, opened 2019-10-05
Fix #120 (handle multiline character literals)

### LINKED ISSUE #120: Failure on multiline character literal
simplecpp produces a failure
~~~
syntax error: No pair for character ("). Can't process file. File is either invalid or unicode, which is currently not supported.
~~~
on
~~~c
const char*ptr="\\
\n";
~~~
.
gcc/cpp and Visuall Studio accept this.

rikardfalkeborn:
Make simplecpp accept the following code:

	const char*ptr="\\
	\n";

Output from simplecpp:

	const char * ptr = "\\n"
	;

Output from gcc -E:

	const char* ptr = "\\n";

Do this by extending the special casing for backslashes to read all
consecutive backslashes before continuing.

This makes it possible to run simplecpp on ftp://ftp.se.debian.org/debian/pool/main/l/lisaac/lisaac_0.39~rc1.orig.tar.gz
(used in cppcheck daca@home). Processing it is a blocker for cppcheck trac ticket [#2435](https://trac.cppcheck.net/ticket/2435).

Fixes #120.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1133,12 +1133,20 @@ std::string simplecpp::TokenList::readUntil(std::istream &istr, const Location &
         backslash = false;
         ret += ch;
         if (ch == '\\') {
-            const char next = readChar(istr, bom);
-            if (next == '\r' || next == '\n') {
-                ret.erase(ret.size()-1U);
-                backslash = (next == '\r');
-            }
-            ret += next;
+            bool update_ch = false;
+            char next = 0;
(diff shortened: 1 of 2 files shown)


danmar:
Feel free to "bump simplecpp" in cppcheck.. now or later. I do that now and then by just copy the files to the cppcheck/externals/simplecpp folder. So in reality we use a dev version of simplecpp in cppcheck.

PR #184, opened 2019-12-06
simplecpp.cpp: Issue error when an explicitly included file is not found

### LINKED ISSUE #183: simplecpp silently ignores when an explicitly included file does not exist
This has been reported in the Cppcheck forum:
https://sourceforge.net/p/cppcheck/discussion/general/thread/f20832cc83/

If `simplecpp` is called with `-include=<file>`, and the file does not exist, there is no error message. `simplecpp` just ignores the file. This could be very confusing for users since the real issue thus can be hidden for a long time.
At [simplecpp:2426](https://github.com/danmar/simplecpp/blob/71adfce5898bad0a059938948b4d9e1ab767189f/simplecpp.cpp#L2426) a not existing file is simply ignored.
GCC aborts with a fatal error in the same case.
To not break the compatibility, `simplecpp` could be changed to print some warning so the user knows that the file is not included but otherwise behave as before.
I suggest the same handling for include files that exist but can not be opened for some reason.

versat:
If a file is explicitly included for example via command line option
`-include=` but can not be opened this error message is issued now.
Example:
$ ./simplecpp -include=blah.h foo.c
int main ( )
{
return 0 ;
}
foo.c:1: missing header: Can not open include file 'blah.h' that is explicitly included for all files.

Fixes #183

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2429,7 +2429,7 @@ std::map<std::string, simplecpp::TokenList*> simplecpp::load(const simplecpp::To
         if (!fin.is_open()) {
             if (outputList) {
                 simplecpp::Output err(fileNumbers);
-                err.type = simplecpp::Output::ERROR;
+                err.type = simplecpp::Output::EXPLICIT_INCLUDE_NOT_FOUND;
                 err.location = Location(fileNumbers);
                 err.msg = "Can not open include file '" + filename + "' that is explicitly included.";
                 outputList->push_back(err);
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2431,7 +2431,7 @@ std::map<std::string, simplecpp::TokenList*> simplecpp::load(const simplecpp::To
                 simplecpp::Output err(fileNumbers);
                 err.type = simplecpp::Output::ERROR;
                 err.location = Location(fileNumbers);
-                err.msg = "Can not open include file '" + filename + "' that is explicitly included for all files.";
+                err.msg = "Can not open include file '" + filename + "' that is explicitly included.";
                 outputList->push_back(err);
             }
             continue;
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2426,8 +2426,16 @@ std::map<std::string, simplecpp::TokenList*> simplecpp::load(const simplecpp::To
             continue;
 
         std::ifstream fin(filename.c_str());
-        if (!fin.is_open())
+        if (!fin.is_open()) {
+            if (outputList) {
+                simplecpp::Output err(fileNumbers);
+                err.type = simplecpp::Output::MISSING_HEADER;
+                err.location = Location(fileNumbers);
+                err.msg = "Can not open include file '" + filename + "' that is explicitly included for all files.";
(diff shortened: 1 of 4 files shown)


simplecpp.cpp
```diff
@@ -2426,8 +2426,16 @@ std::map<std::string, simplecpp::TokenList*> simplecpp::load(const simplecpp::To
             continue;
 
         std::ifstream fin(filename.c_str());
-        if (!fin.is_open())
+        if (!fin.is_open()) {
+            if (outputList) {
+                simplecpp::Output err(fileNumbers);
+                err.type = simplecpp::Output::MISSING_HEADER;
```
versat: Should I better use `simplecpp::Output::ERROR` here?

versat:
Hmm, while this works for `simplecpp`, this is not a solution for Cppcheck.
In Cppcheck `outputList` is `NULL`, so it does not make any difference.

simplecpp.cpp
versat: After looking at how Cppcheck handles the `outputList` I think it does make more sense to use `ERROR` here. I have changed it.

simplecpp.cpp
```diff
@@ -2426,8 +2426,16 @@ std::map<std::string, simplecpp::TokenList*> simplecpp::load(const simplecpp::To
             continue;
 
         std::ifstream fin(filename.c_str());
-        if (!fin.is_open())
+        if (!fin.is_open()) {
+            if (outputList) {
+                simplecpp::Output err(fileNumbers);
+                err.type = simplecpp::Output::ERROR;
+                err.location = Location(fileNumbers);
+                err.msg = "Can not open include file '" + filename + "' that is explicitly included for all files.";
```
danmar: Unless you can motivate "for all files" please remove that. Imagine this TU has 1 source file and 4 headers that are included. I do not believe that the file is included 5 times in this TU.

simplecpp.cpp
```diff
@@ -2426,8 +2426,16 @@ std::map<std::string, simplecpp::TokenList*> simplecpp::load(const simplecpp::To
             continue;
 
         std::ifstream fin(filename.c_str());
-        if (!fin.is_open())
+        if (!fin.is_open()) {
+            if (outputList) {
+                simplecpp::Output err(fileNumbers);
+                err.type = simplecpp::Output::ERROR;
```
danmar: Please add a new enum constant. The `ERROR` constant is used when `#error` is encountered in the "active" code.

versat: I see, I will add a new enum constant.

simplecpp.cpp
versat: You are right, I meant only source files so the message is not correct. Maybe it is best to remove "for all files".

versat:
Sorry, I had no better idea than `EXPLICIT_INCLUDE_NOT_FOUND` as new error type. Do you have any better idea? I have thought about `PARAMETER_ERROR` or something like that, but I am not sure if that is too generic or not.

danmar:
> Do you have any better idea?

Not directly. Let's use this for now.. it's ok to change it later.

PR #186, opened 2019-12-23
Fixed -Wshadow warning.

orbitcowboy:
The local variable `altop` was shadowed by the global at https://github.com/danmar/simplecpp/blob/c27e5d46dced1f8ead1f3a755ce539cdd0ee42b2/simplecpp.cpp#L2241

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1020,15 +1020,15 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
 {
     Token * const tok1 = tok;
     for (const char *op = "&^|"; *op; op++) {
-        const std::string* alTop;
+        const std::string* alternativeOp;
         if (*op == '&')
-            alTop = &BITAND;
+            alternativeOp = &BITAND;
         else if (*op == '|')
-            alTop = &BITOR;
+            alternativeOp = &BITOR;
         else
-            alTop = &XOR;
+            alternativeOp = &XOR;
         for (tok = tok1; tok && tok->op != ')'; tok = tok->next) {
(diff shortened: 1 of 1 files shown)


simplecpp.cpp
```diff
@@ -1020,15 +1020,15 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
 {
     Token * const tok1 = tok;
     for (const char *op = "&^|"; *op; op++) {
-        const std::string* altop;
```
danmar: I believe `altop` is a abbreviation for `alternative operator` or something like that.

orbitcowboy: I will update this in approx. two weeks. I suppose "alternativeOp" could be used as local variable name. Ok?

orbitcowboy:
> hmm I think `Wshadow` is valuable basically.
> 
> However I am not sure that it should be fixed like this. If a `t` is written instead of `T` somewhere by mistake then bad things will happen.
> 
> It would be preferable if the variable names can be distinguished better. I don't have a suggestion off the top of my head.
> 
> Maybe it's better that the global variable is not a abbreviation. Using abbreviations for global variables should be avoided imho unless it's 100% clear for everybody what it means.

Just to be clear. This is not triggered by a GCC -Wshadow warning. I stumbled over this during debugging. Shouldn't Cppcheck detect this issue as well?

simplecpp.cpp
danmar: that is ok

danmar:
> Shouldn't Cppcheck detect this issue as well?

yes I think so

PR #190, opened 2020-05-15
Add github action: CI windows

amai2012:


danmar:
looks good.

I wonder can we replace some appveyor jobs with github actions now? That would be faster right?

amai2012:
Yes, currently github action is much faster

PR #195, opened 2020-10-01
Create simple.cpp

ashwinisf:


danmar:
Thanks! But I do not see how this helps.

PR #205, opened 2021-01-13
Fix preprocessing of several ##

Ken-Patrick:
```
#define MAX_FOO (2 * 1)
#define MAX_FOO_AA (3 * 1)

#define M(UpperCaseName, b)                \
  do {                                     \
    int MaxValue = MAX_##UpperCaseName;    \
    if (b) {                               \
      MaxValue = MAX_##UpperCaseName##_AA; \
    }                                      \
  } while (0)

static void f(bool b) { M(FOO, b); }
```

was a preprocessor error
`a.cpp:4:0: error: failed to expand 'M', Invalid ## usage when expanding 'M'. [preprocessorErrorDirective]`
because `MAX_##UpperCaseName##_AA` was expanded to (2 * 1)##_AA, while gcc or clang happily expand it to `(3 * 1)`.

I don't really know how to add a test here (I've quickly tested with cppcheck and it seemed ok). I see there are some tests in cppcheck/test/testpreprocessor.cpp but I'm probably a bit confused. Any help is more than welcome here ;-)

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1880,7 +1880,6 @@ namespace simplecpp {
             }
 
             const Token *nextTok = B->next;
-
             if (varargs && tokensB.empty() && tok->previous->str() == ",")
                 output->deleteToken(A);
             else if (strAB != "," && macros.find(strAB) == macros.end()) {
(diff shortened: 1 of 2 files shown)


danmar:
please add some `hashhash` test: https://github.com/danmar/simplecpp/blob/master/test.cpp#L1975

Ken-Patrick:
> please add some `hashhash` test: https://github.com/danmar/simplecpp/blob/master/test.cpp#L1975

Of course. (somehow, I hadn't spotted the file, thanks for pointing it out !)

danmar:
some jobs failed.. maybe it works if I restart..

Ken-Patrick:
> some jobs failed.. maybe it works if I restart..

I tried, but no luck. IIUC, there's an issue with the macos11 workers of the CI, they don't run and fail (I tried with an empty commit on master, and got a similar failure)...

danmar:
I will retry.. it would feel good to have that green. probably it's a temporary problem.

danmar (approved):

PR #213, opened 2021-03-15
Linemarkers

silver-popov:
Add linemarkers https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2624,11 +2624,14 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
               // Add linemarker
               // https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html
               std::string linemarker = "# ";
-              linemarker += std::to_string(rawtok->location.line);
+              Location loc = rawtok->location;
+              --loc.line;
+              loc.col = 0;
+              linemarker += std::to_string(loc.line);
               linemarker += " \"";
-              linemarker += rawtok->location.file ();
+              linemarker += loc.file ();
               linemarker += "\" 2";
-              output.push_back (new Token (linemarker, rawtok->location));
+              output.push_back (new Token (linemarker, loc));
             }
             continue;
         }
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2623,15 +2623,16 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
             if (rawtok != rawtokens.cfront ()) {
               // Add linemarker
               // https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html
-              std::string linemarker = "# ";
               Location loc = rawtok->location;
               --loc.line;
               loc.col = 1;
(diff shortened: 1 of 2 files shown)


simplecpp.cpp
```diff
@@ -2620,6 +2620,20 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
         if (rawtok == NULL) {
             rawtok = includetokenstack.top();
             includetokenstack.pop();
+            if (rawtok && rawtok != rawtokens.cfront ()) {
+              // Add linemarker
```
danmar, line 2624: I want that you only add the line markers in `stringify` output. The line markers are only for the debug output.

People that read the raw tokens from simplecpp have all location info they need without these markers.

silver-popov: I've tried to do it. But there is not enough information in the final token list.
So I was forced to implement it at the preprocessing stage.

danmar: how about this implementation:
```
std::string simplecpp::TokenList::stringify() const
{
    std::ostringstream ret;
    Location loc(files);
    for (const Token *tok = cfront(); tok; tok = tok->next) {
        if (tok->location.line != loc.line || tok->location.fileIndex != loc.fileIndex) {
            ret << "\n# " << tok->location.line << " \"" << tok->location.file() << "\"\n";
            loc = tok->location;
        }

        while (tok->location.line > loc.line) {
            ret << '\n';
            loc.line++;
        }

        if (sameline(tok->previous, tok))
            ret << ' ';

        ret << tok->str();

        loc.adjust(tok->str());
    }

    return ret.str();
}
```

silver-popov: First, I wanted to base all the include logic on the #line directive or other information in the output token list. But it is impossible.

Let's consider the example:
// Test_I3.idl
#include "Test_I1.idl"
#include "Test_I2.idl"
...bla-bla3

// Test_I1.idl
...bla-bla1

// Test_I2.idl
...bla-bla2

I need to produce the following output:
// Test_I3.h
#include "Test_I1.h
#include "Test_I2.h
bla-bla_cpp3

When I first meet the token from bla-bla2, the tok->location.fileIndex points to Test_I1.idl. Not to Test_I3.idl. Because there are no tokens between #include "Test_I1.idl" and #include "Test_I2.idl" were intervening the #line directive. So I can not detect that Test_I2.idl was included in Test_I3.idl file. 
In other words, there is no difference in simplecpp output tokens when Test_I2.idl was included: in Test_I3.idl, or in Test_I1.idl. In both cases, the output tokens will be the same.

danmar: quick comment.. I do not want to have such line marker tokens in the token list this is just some debug output.. so well try to embed the information in Token or TokenList or something..

danmar: > In other words, there is no difference in simplecpp output tokens when Test_I2.idl was included: in Test_I3.idl, or in Test_I1.idl. In both cases, the output tokens will be the same.

that is not information that I normally care about but feel free to store that info somehow..

Issue #214, opened 2021-04-22
Escape sequences in character literals are not handled.


keinflue:
```
#if '\'' != '\\'
ok
#else
fail
#endif
```
This results in
```



fail
```

when it should result in `ok`.

It seems that escape sequences in character literals are not handled at all and instead they are interpreted as backslash literal.

If the contribution would be welcome, I could try my hand at implementing support at least partially. It seems `simplifyNumbers` just requires special casing for the possible escape sequences.

danmar:
> If the contribution would be welcome, I could try my hand at implementing support at least partially. It seems simplifyNumbers just requires special casing for the possible escape sequences.

Yes your contribution would be highly welcome.

keinflue:
This issue was resolved with #216.

PR #215, opened 2021-04-22
Partial fix for character values with escape sequences (#214)

keinflue:
The first commit is a minimal fix for the evaluation of character literals with escape sequences in the most common cases.

It does not fix the issue for

1. universal character name escape sequences
2. prefixed character literals (L'...', u'...', U'...', u8'....')
3. the `\e` GNU extension

I made the use of an unsupported escape sequence, including 1. and 3., fatal. I am not sure whether this is preferred over an incorrect integer representation.

For 2. a wrong representation is still used. I intended to partially implement 2., but it seems that the prefixed literals never make it to `simplifyNumbers`, so it might be more complicated.

For multicharacter literals all but the first one are ignored. That doesn't match the behavior of typical implementations, but is, I think, standard-conform.

Following the code already present, character values are determined by the host execution character set and wrapped/truncated to the `0` - `255` range. This implies an unsigned `char` type. I am not sure whether/how I should take `char` signedness into account. It might make sense to make out-of-range values fatal instead.

The second commit makes conditions in which an expected number can not be read fatal. I introduced it to make `\x` without any digits or too many digits fatal. Again I am not sure whether this is preferred and this change in particular might have effects that I didn't anticipate.

TODO: Add test cases.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2309,8 +2309,66 @@ static void simplifyNumbers(simplecpp::TokenList &expr)
             continue;
         if (tok->str().compare(0,2,"0x") == 0)
             tok->setstr(toString(stringToULL(tok->str())));
-        else if (tok->str()[0] == '\'')
-            tok->setstr(toString(tok->str()[1] & 0xffU));
+        else if (tok->str()[0] == '\'') {
+            // this is a character literal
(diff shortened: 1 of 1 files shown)


danmar:
why did you close it? if you want that I review it then please open it again..

PR #219, opened 2021-05-06
Support some more GCC extensions for character literals

keinflue:
This fixes daca reports such as
```
tracetuner_3.0.6beta/src/mkchk/checkbcphd.c:0:0: error: Internal Error. MathLib::toLongNumber: characterLiteralToLL('\%') => invalid escape sequence [cppcheckError]
xview-3.2p1.4/util/xgettext/xgettext.c:0:0: error: Internal Error. MathLib::toLongNumber: characterLiteralToLL('\(') => invalid escape sequence [cppcheckError]
afterstep-devel-2.2.12/libAfterConf/Color.c:0:0: error: Internal Error. MathLib::toLongNumber: characterLiteralToLL('\{') => invalid escape sequence [cppcheckError]
NetworkManager-1.30.0/src/core/settings/plugins/ifcfg-rh/shvar.c:0:0: error: Internal Error. MathLib::toLongNumber: characterLiteralToLL('\E') => invalid escape sequence [cppcheckError]
```
which use extensions that are supported by both GCC and Clang.

This does not fix the following report which is also accepted by GCC, but not Clang:
```
sparse-0.6.3/validation/escapes.c:0:0: error: Internal Error. MathLib::toLongNumber: characterLiteralToLL('\x7890') => numeric escape sequence too large [cppcheckError]
```
The behavior of this literal in GCC is a bit weird, resulting in `-0x70` if `char` is signed and `0x90` otherwise. According to the standard it is a contraint violation. I suppose it is more likely that this is written unintentionally, so I think an error message isn't too bad.

Both Clang and GCC also accept invalid literals like `'\z'` with a warning, interpreting them by simply dropping the backslash. I did not add this behavior, because there wasn't any daca report for them so far.

I think all of these extensions should result in a portability report, but I don't see any mechanism to signal that to the caller of `characterLiteralToLL` at the moment.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2402,6 +2402,12 @@ long long simplecpp::characterLiteralToLL(const std::string& str)
                 throw std::runtime_error("unexpected end of character literal");
 
             switch (escape) {
+            // obscure GCC extensions
+            case '%':
+            case '(':
+            case '[':
+            case '{':
+            // standard escape sequences
             case '\'':
             case '"':
             case '?':
@@ -2431,8 +2437,9 @@ long long simplecpp::characterLiteralToLL(const std::string& str)
                 value = static_cast<unsigned char>('\v');
                 break;
 
-            // ESC extension
+            // GCC extension for ESC character
             case 'e':
+            case 'E':
                 value = static_cast<unsigned char>('\x1b');
                 break;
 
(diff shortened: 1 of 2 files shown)


danmar:
> I think all of these extensions should result in a portability report, but I don't see any mechanism to signal that to the caller of characterLiteralToLL at the moment.

we can think about that. 

if we want that a portability warning is written during preprocessing then the best way would be to add a warning to the `outputList`..

if we want that cppcheck writes a portability warning during analysis.. then something different is wanted.

PR #220, opened 2021-05-08
Basic support for UTF-8 source in character literals

keinflue:
This should resolve most daca reports of the following kinds:
```
head glibc-2.31/libio/tst-ftell-partial-wide.c:0:0: error: Internal Error. MathLib::toLongNumber: characterLiteralToLL(L'\343\201\224') => non-ASCII source characters supported only in narrow character literals [cppcheckError]
head foot-1.6.4/main.c:0:0: error: Internal Error. MathLib::toLongNumber: characterLiteralToLL(U'\303\266') => non-ASCII source characters supported only in narrow character literals [cppcheckError]
head icu/source/i18n/formatted_string_builder.cpp:0:0: error: Internal Error. MathLib::toLongNumber: characterLiteralToLL(u'') => non-ASCII source characters supported only in narrow character literals [cppcheckError]
```
by assuming that non-narrow literals are passed to `characterLiteralToLL` encoded in UTF-8. This is in contrast with the handling of narrow literals, for which we assume a single-byte ASCII-compatible encoding for compatibility with source files in non-unicode character sets.

Ideally the source file is correctly decoded earlier in the process, but this should give correct results in the most common cases. It may however result in incorrect values if the assumed encodings are not correct.

I am not sure whether adding this extra complexity to simplecpp itself is worth it, especially since it is just a half-measure to the problem, but I felt like trying to write this and so I am suggesting it now.

As suggested in other discussion, an alternative would be to simply not assign values in these cases, continue failing if this happens during preprocessing (unlikely) and for integration in cppcheck signal an invalid value, since cppcheck isn't reliant on all of the values being present, but I don't know how I would implement this signalling in cppcheck.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2334,10 +2334,11 @@ static unsigned long long stringToULLbounded(
 /* Converts character literal (including prefix, but not ud-suffix)
  * to long long value.
  *
- * Assumes ASCII-compatible single-byte encoded str.
+ * Assumes ASCII-compatible single-byte encoded str for narrow literals
+ * and UTF-8 otherwise.
  *
  * For target assumes
- * - UTF-8 execution character set encoding or encoding matching str
+ * - execution character set encoding matching str
  * - UTF-32 execution wide-character set encoding
  * - requirements for __STDC_UTF_16__, __STDC_UTF_32__ and __STDC_ISO_10646__ satisfied
  * - char16_t is 16bit wide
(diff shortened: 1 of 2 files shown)


keinflue:
@danmar It seems to me that http://cppcheck1.osuosl.org:8000/head-cppcheckUnhandledChar does not actually show the reports. For example there is a `'\c'` literal for version 2.4 reported in http://cppcheck1.osuosl.org:8000/sparse which should still be reported in head, but isn't. Looking over the donate-cpu scripts it seems that `information` severity is handled differently and saved to a different file, but I don't see how to get these reports from the server.

Also, unrelated, I was trying to get a trac account by asking in the IRC channel to report some cppcheck false positives I found, but haven't gotten a response so far. Should I just keep asking there or what is the procedure?

danmar:
> Looking over the donate-cpu scripts it seems that information severity is handled differently and saved to a different file, but I don't see how to get these reports from the server.

ah ok.. I need to investigate. I assume there is a way..

> unrelated, I was trying to get a trac account by asking in the IRC channel to report some cppcheck false positives I found

somehow send me a htpasswd hash and I will create an account for you. you can generate it online if you want, google "htpasswd generator"..

danmar:
> It seems to me that http://cppcheck1.osuosl.org:8000/head-cppcheckUnhandledChar does not actually show the reports.

hmm.. I have changed the severity to portability. Hope we will only get warnings when there are non-standard character literals.

I hope results will be shown here:

http://cppcheck1.osuosl.org:8000/head-nonStandardCharLiteral

PR #221, opened 2021-05-08
Improve validation of UTF-8 encoded character literals

keinflue:
Rejects overlong encodings and adds some more rejection test cases.

Does not handle noncharacters specifically. If the noncharacter code
points are going to be used for program-internal purposes, this decoder
needs to be updated to reject or replace them on incoming input.


danmar:
For information.. if there are changes in simplecpp that I want to have in cppcheck I just "bump simplecpp"..
so if you feel that you are "done" and want to test then feel free to "bump simplecpp" (just copy simplecpp.cpp and simplecpp.h to cppcheck/externals/simplecpp)

PR #232, opened 2022-01-27
fixed -Wzero-as-null-pointer-constant Clang warnings

firewave:
This adds `nullptr` as a define when in c++03 mode allowing the code to be compiled against c++11 code without raising such warnings. AppleClang always defines it - even in c++03 mode.


danmar:
I think it's ok to replace 0 with nullptr.

firewave:
I took care of the "hanging" ubuntu-16.04 jobs in #233

PR #233, opened 2022-01-27
small CI improvements - always run all jobs and removed obsolete ubuntu-16.04

firewave:


firewave:
@danmar Please merge this first as it will prevent the unavailable CI runners from hanging for 24(!) hours.

danmar:
I like failfast. If something fails I want that it is clear where it fails.

firewave:
> I don't understand how fail-fast can cause hang. but if it just prevents that other jobs are terminated I think it's ok.

The hang is stupid GitHub behavior and not related to `fail-fast` (hence the two commits). We are specifying a runner which no longer exists and then it waits a day for such a runner to appear.

PR #234, opened 2022-01-27
main.cpp: added -q (quiet mode)

firewave:


main.cpp
```diff
@@ -49,6 +53,8 @@ int main(int argc, char **argv)
         std::cout << "  -IPATH          Include path." << std::endl;
         std::cout << "  -include=FILE   Include FILE." << std::endl;
         std::cout << "  -UNAME          Undefine NAME." << std::endl;
+        std::cout << "  -std=STD        Specify standard." << std::endl;
+        std::cout << "  -q              Quiet mode (no output)." << std::endl;
```
danmar, line 57: why would you run it in quiet mode if there is no output?

firewave: Profiling (less code being executed) and crash testing.

main.cpp
```diff
@@ -36,6 +37,9 @@ int main(int argc, char **argv)
                 if (std::strncmp(arg, "-std=",5)==0)
                     dui.std = arg + 5;
                 break;
+			case 'q':
```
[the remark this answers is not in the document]
firewave: done

main.cpp
firewave: Also if we would change the exitcode based on if the parsing was successful or not it could use used as a command-line tool and in that case you also might not want *all* the output.

PR #240, opened 2022-02-20
consider cpp conditional expression tokens for macro usage

josecm:
This should be done even if such macros might only be defined further ahead.
Specifically, to mark include guard macros as used.

This fixes misra rule 2.5 false positives for the misra cppcheck addon
reported in cppcheck trac ticket #10751.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2912,6 +2912,8 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
             includetokenstack.push(f->second->cfront());
     }
 
+    std::map<std::string, std::list<Location> > maybeUsedMacros;
+
     for (const Token *rawtok = nullptr; rawtok || !includetokenstack.empty();) {
         if (rawtok == nullptr) {
             rawtok = includetokenstack.top();
@@ -3077,11 +3079,13 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
                 bool conditionIsTrue;
                 if (ifstates.top() == ALWAYS_FALSE || (ifstates.top() == ELSE_IS_TRUE && rawtok->str() != ELIF))
                     conditionIsTrue = false;
-                else if (rawtok->str() == IFDEF)
+                else if (rawtok->str() == IFDEF) {
                     conditionIsTrue = (macros.find(rawtok->next->str()) != macros.end() || (hasInclude && rawtok->next->str() == HAS_INCLUDE));
-                else if (rawtok->str() == IFNDEF)
+                    maybeUsedMacros[rawtok->next->str()].push_back(rawtok->next->location);
+                } else if (rawtok->str() == IFNDEF) {
                     conditionIsTrue = (macros.find(rawtok->next->str()) == macros.end() && !(hasInclude && rawtok->next->str() == HAS_INCLUDE));
-                else { /*if (rawtok->str() == IF || rawtok->str() == ELIF)*/
+                    maybeUsedMacros[rawtok->next->str()].push_back(rawtok->next->location);
+                } else { /*if (rawtok->str() == IF || rawtok->str() == ELIF)*/
                     TokenList expr(files);
                     for (const Token *tok = rawtok->next; tok && tok->location.sameline(rawtok->location); tok = tok->next) {
                         if (!tok->name) {
@@ -3094,6 +3098,7 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
                             const bool par = (tok && tok->op == '(');
                             if (par)
                                 tok = tok->next;
+                            maybeUsedMacros[rawtok->next->str()].push_back(rawtok->next->location);
                             if (tok) {
                                 if (macros.find(tok->str()) != macros.end())
                                     expr.push_back(new Token("1", tok->location));
@@ -3146,6 +3151,8 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
                             }
                             continue;
                         }
+
+                        maybeUsedMacros[rawtok->next->str()].push_back(rawtok->next->location);
 
                         const Token *tmp = tok;
                         if (!preprocessToken(expr, &tmp, macros, files, outputList)) {
@@ -3256,7 +3263,9 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
     if (macroUsage) {
         for (simplecpp::MacroMap::const_iterator macroIt = macros.begin(); macroIt != macros.end(); ++macroIt) {
             const Macro &macro = macroIt->second;
-            const std::list<Location> &usage = macro.usage();
+            std::list<Location> usage = macro.usage();
+            const std::list<Location>& temp = maybeUsedMacros[macro.name()];
+            usage.insert(usage.end(), temp.begin(), temp.end());
             for (std::list<Location>::const_iterator usageIt = usage.begin(); usageIt != usage.end(); ++usageIt) {
                 MacroUsage mu(usageIt->files, macro.valueDefinedInCode());
                 mu.macroName = macro.name();


simplecpp.cpp
```diff
@@ -2842,6 +2842,14 @@ static bool preprocessToken(simplecpp::TokenList &output, const simplecpp::Token
     return true;
 }
 
+static void addMaybeUsedMacro(std::map<std::string, std::list<simplecpp::Location> > &macros, const simplecpp::Token *tok) {
+    if(tok == nullptr) return;
+    if(macros.count(tok->str()) == 0) {
```
danmar, line 2847: as far as I see you can replace this:
```
    if(macros.count(tok->str()) == 0) {
        macros.insert(std::make_pair(tok->str(), std::list<simplecpp::Location>()));
    }
    macros.find(tok->str())->second.push_back(tok->location);
```
with:
```
    macros[tok->str()].push_back(tok->location);
```

simplecpp.cpp
```diff
@@ -2842,6 +2842,14 @@ static bool preprocessToken(simplecpp::TokenList &output, const simplecpp::Token
     return true;
 }
 
+static void addMaybeUsedMacro(std::map<std::string, std::list<simplecpp::Location> > &macros, const simplecpp::Token *tok) {
```
danmar, line 2845: well.. in general I prefer to pass by address instead of pass by reference for output parameters. When reading the simplecpp code and you see a function call.. if there is a `&` in front of a parameter in a function call that generally means the parameter is written.

danmar (commented):
good

simplecpp.cpp
josecm: That makes sense. I am not that used to programming in C++, so I am not really familiar with the idioms. Let me fix that.

simplecpp.cpp
josecm: Sorry, once again: not really familiar with C++. Well, if this is the case, not really that worth it having the helper function. Let me refactor this.

simplecpp.cpp
josecm: I completely remove the helper function, given that it becomes a one-liner by your first suggestion.

danmar:
I am not sure that this is a proper fix for trac ticket #10751.. imagine:

test.c:
```
#include <stdio.h>
#ifndef DATA
#define DATA
#endif
int x;
```
This is not a include guard. DATA is unused so I would guess that the misra addon should warn about it.

I would suggest that the addon tries to determine if a macro is a include guard.

For instance; If there is code above the `#ifndef` or below the `#endif` then it's not a include guard. The reasonable implementation is to use `#ifndef` so we can assume that `#ifdef` is not a include guard.

josecm:
My approach was exactly *not* to target specifically include guards. I looked at it from a more broad perspective. That is, I considered that using a macro in a pre-processor conditional would constitute a macro usage (even if the macro is defined further ahead). This in turn solves the issue with misra rule 2.5 false positives.

However, if you do want to just target the include guard exception specifically,  I ask you what would be the best approach and where in the code should the modifications go? I noticed that the getConfigs in lib/preprocessor.cpp already detects include guards. I would I go about passing this information to the misra python addon?

danmar:
> I ask you what would be the best approach and where in the code should the modifications go? I noticed that the getConfigs in lib/preprocessor.cpp already detects include guards. I would I go about passing this information to the misra python addon?

yes however it's clumpsy to add all Cppcheck information in the dump file.

I believe the dumpfile has the information we need..

```
  <directivelist>
    <directive file="m25.c" linenr="1" str="#include &lt;stdio.h&gt;"/>
    <directive file="m25.c" linenr="2" str="#include &quot;1.h&quot;"/>
    <directive file="m25.c" linenr="3" str="#ifndef DATA"/>
    <directive file="m25.c" linenr="4" str="#define DATA 1"/>
    <directive file="m25.c" linenr="5" str="#endif"/>
    <directive file="1.h" linenr="1" str="#ifndef __1_H__"/>
    <directive file="1.h" linenr="2" str="#define __1_H__"/>
    <directive file="1.h" linenr="4" str="#endif"/>
  </directivelist>
  <tokenlist>
    <token id="0x55572e291cb0" file="1.h" linenr="3" column="1" str="int" scope="0x55572e286370" type="name"/>
    <token id="0x55572e28cdb0" file="1.h" linenr="3" column="5" str="x1" scope="0x55572e286370" type="name" varId="1" variable="0x55572e285ab0" valueType-type="int" valueType-sign="signed" valueType-reference="None"/>
    <token id="0x55572e28ce10" file="1.h" linenr="3" column="7" str=";" scope="0x55572e286370"/>
    <token id="0x55572e291e50" file="m25.c" linenr="6" column="1" str="int" scope="0x55572e286370" type="name"/>
    <token id="0x55572e28c6d0" file="m25.c" linenr="6" column="5" str="x" scope="0x55572e286370" type="name" varId="2" variable="0x55572e286640" valueType-type="int" valueType-sign="signed" valueType-reference="None"/>
    <token id="0x55572e28c7c0" file="m25.c" linenr="6" column="6" str=";" scope="0x55572e286370"/>
  </tokenlist>
```
the __1_H__ should be classified as a include guard here. in the directive list you can see what the first and last directives are in each file. if the first directive is a ifndef and the last directive is the corresponding endif then that is a candidate. in the tokenlist you can determine what the first line of code and last line of code is in the file. if all the code is enclosed inside the  `ifndef`-`endif` scope and __1_H__ is also defined in that scope.. I believe we can classify __1_H__ is a include guard.

josecm:
I'll take a look at how to do this based on your suggestion. However, I would first ask you to reconsider my first approach in light of the guideline text itself and [this misra forum thread](https://forum.misra.org.uk/thread-1317.html) (pointed out earlier by Henrik Nyholm in the [sourceforge discussion](https://sourceforge.net/p/cppcheck/discussion/general/thread/4d9f3a5013/)).

Rule 2.5 only directly addresses unused macros, it does not explicitly open an exception for include guards. Include guards are mentioned in a completely separate guideline (directive 4.10) which suggests two forms of include guards: one using the classic `#ifndef MACRO`,  other using  `#if !defined (MACRO)`.

Technically, rule 2.5 does not have any awareness of include guards. Nevertheless, they clearly must not be flagged as a violation. Is this an exception or part of the rule definition itself? IMO, the latter is correct. So, I believe that in order to correctly implement the checker for this rule, the central question becomes what exactly constitutes a macro usage. More specifically, if macros used in any pre-processor conditional expression should be considered as a used macro. Although I could not find any defintion of macro usage in the guideline document, the question is indirectly answered by what appears to be an official misra representative in the aforementioned misra forums thread: "The #ifndef HEADER_H is considered to be a use of the macro name, even though it appears before the #define."

From this I conclude that even in your previous example, the macro should be considered as used.

danmar:
> From this I conclude that even in your previous example, the macro should be considered as used.

ok I agree. The your approach here is better..

simplecpp.cpp
```diff
@@ -3257,12 +3264,20 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
         for (simplecpp::MacroMap::const_iterator macroIt = macros.begin(); macroIt != macros.end(); ++macroIt) {
             const Macro &macro = macroIt->second;
             const std::list<Location> &usage = macro.usage();
-            for (std::list<Location>::const_iterator usageIt = usage.begin(); usageIt != usage.end(); ++usageIt) {
-                MacroUsage mu(usageIt->files, macro.valueDefinedInCode());
-                mu.macroName = macro.name();
-                mu.macroLocation = macro.defineLocation();
-                mu.useLocation = *usageIt;
-                macroUsage->push_back(mu);
+            std::list<std::list<Location> > usageLists;
```
danmar, line 3267: hmm.. an alternative could be to append the two lists into one list..

    std::list<Location> usage = macro.usage();
    const std::list<Location>& temp = maybeUsedMacros[macro.name()];
    usage.insert(usage.end(), temp.begin(), temp.end());

firewave:
> I would suggest that the addon tries to determine if a ifndef block is a include guard.

(I am 100% not sure about this) but we might need logic to detect (and handle) include guards to solve https://trac.cppcheck.net/ticket/10761 (I have not looked into this potential issue at all).

danmar:
@firewave in my humble opinion if there is no include guard then we should preprocess the header every time the header is re-included. otherwise the preprocessor output and therefore analysis will be wrong.

if there are preprocessor statements and we don't handle them properly we should of course fix that but that fix should not be specific about include guards then.

firewave:
> @firewave in my humble opinion if there is no include guard then we should preprocess the header every time the header is re-included. otherwise the preprocessor output and therefore analysis will be wrong.
> 
> if there are preprocessor statements and we don't handle them properly we should of course fix that but that fix should not be specific about include guards then.

Okay, let's not get off-topic here. I will add an example to the ticket soon and we can discuss it then.

simplecpp.cpp
josecm: I was trying to avoid deep copies, but agree the code became too convoluted. Once again my C++ inexperience is showing. I'll apply your suggestion.

josecm: Force pushed!

josecm:
> ok I agree. Then your approach here is better..

Thanks for reconsidering. I don't think it is better or worse, I was just trying to interpret misra guidelines as strictly as possible, to implement the checker correctly. Actually, it was my [discussion with Henrik](https://sourceforge.net/p/cppcheck/discussion/general/thread/4d9f3a5013/) that led me to conclude that was the correct intrepretation.

PR #242, opened 2022-02-24
exposed -std to define mapping in interface / add support for more -std values / added setting of __STDC_VERSION__

firewave:
This exposes logic which is duplicated in Cppcheck.

Actually Cppcheck doesn't even need this logic (the code which uses this and was added by me was just the wrong approach) but that's a later cleanup within Cppcheck. And it still makes sense to have this functionality.


danmar:
lgtm

PR #243, opened 2022-03-04
only stringify the last line in `TokenList::readfile()` if necessary 

firewave:
`lastLine()` is only used to check for preprocessor directives. So if it isn't such there's no need to stringify it saving a lot of unnecessary `std::string` operations.

The intermediate increase comes from not constructing `std::string` objects on-the-fly for `std::string::insert()` and simply using the literals.

Testing with `-q -Ilib/ -D__GNUC__ lib/valueflow.cpp` this saves about 8% of total Ir.

Clang 13 `298,836,106` -> `295,950,055` -> `267,545,621`
GCC 11 `305,058,160` -> `299,749,788` -> `271,347,737`

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -509,6 +509,8 @@ void simplecpp::TokenList::readfile(std::istream &istr, const std::string &filen
 
             if (oldLastToken != cback()) {
                 oldLastToken = cback();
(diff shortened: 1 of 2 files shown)


firewave:
This could probably be simplified by getting rid of `lastLine()` completely and creating a matcher function instead.

PR #245, opened 2022-03-07
Improve support for __has_include

ceggers-arri:
Fixes:

> error: failed to evaluate #if condition [preprocessorErrorDirective]
> #if __has_include(<span>)
> ^

Please update simplecpp in cppcheck.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -3153,17 +3153,30 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
                             const bool par = (tok && tok->op == '(');
                             if (par)
                                 tok = tok->next;
+                            bool closingAngularBracket = false;
                             if (tok) {
                                 const std::string &sourcefile = rawtok->location.file();
-                                const bool systemheader = (tok->str()[0] == '<');
-                                const std::string header(realFilename(tok->str().substr(1U, tok->str().size() - 2U)));
+                                const bool systemheader = (tok && tok->op == '<');
+                                std::string header;
+
+                                if (systemheader) {
+                                    while ((tok = tok->next) && tok->op != '>')
+                                        header += tok->str();
+                                    header = realFilename(header);
+                                    if (tok && tok->op == '>')
+                                        closingAngularBracket = true;
+                                }
+                                else {
+                                    header = realFilename(tok->str().substr(1U, tok->str().size() - 2U));
+                                    closingAngularBracket = true;
+                                }
                                 std::ifstream f;
                                 const std::string header2 = openHeader(f,dui,sourcefile,header,systemheader);
                                 expr.push_back(new Token(header2.empty() ? "0" : "1", tok->location));
                             }
                             if (par)
                                 tok = tok ? tok->next : nullptr;
-                            if (!tok || !sameline(rawtok,tok) || (par && tok->op != ')')) {
+                            if (!tok || !sameline(rawtok,tok) || (par && tok->op != ')') || (!closingAngularBracket)) {
                                 if (outputList) {
                                     Output out(rawtok->location.files);
                                     out.type = Output::SYNTAX_ERROR;
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2341,6 +2341,63 @@ static void simplifySizeof(simplecpp::TokenList &expr, const std::map<std::strin
     }
 }
 
+/** Evaluate __has_include(file) */
+static std::string openHeader(std::ifstream &f, const simplecpp::DUI &dui, const std::string &sourcefile, const std::string &header, bool systemheader);
+static void simplifyHasInclude(simplecpp::TokenList &expr, const simplecpp::DUI &dui)
+{
+    for (simplecpp::Token *tok = expr.front(); tok; tok = tok->next) {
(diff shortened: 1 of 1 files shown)


simplecpp.cpp
```diff
@@ -3128,17 +3128,25 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
...
                             if (par)
                                 tok = tok->next;
+                            const bool systemheader = (tok && tok->op == '<');
+                            if (systemheader)
+                                tok = tok->next;
                             if (tok) {
                                 const std::string &sourcefile = rawtok->location.file();
-                                const bool systemheader = (tok->str()[0] == '<');
-                                const std::string header(realFilename(tok->str().substr(1U, tok->str().size() - 2U)));
+                                const std::string header(realFilename(systemheader ? tok->str() : (tok->str().substr(1U, tok->str().size() - 2U))));
```
danmar, line 3136: unfortunate there are not proper unit tests for this. I get the feeling that with the old code there could be a token like this: `<foo.h>` with your code you seem to assume that the tokens will be `<` , `foo.h` and `>`..
it is a while since this was written so I am not sure.

danmar (commented):
could you double check what tokens there are for `<foo.h>` please

simplecpp.cpp
ceggers-arri: The ```/``` character in ```<foo/foo.h>``` also didn't work...

Additionally I added support for evaluation of
```#define FOO __has_include(<foo/foo.h>)```

File: simplecpp.cpp [hunk 2]
```diff
@@ -2341,6 +2341,63 @@ static void simplifySizeof(simplecpp::TokenList &expr, const std::map<std::strin
...
+static void simplifyHasInclude(simplecpp::TokenList &expr, const simplecpp::DUI &dui)
+{
+    for (simplecpp::Token *tok = expr.front(); tok; tok = tok->next) {
+        if (tok->str() != "__has_include")
+            continue;
+        simplecpp::Token *tok1 = tok->next;
+        if (!tok1) {
+            throw std::runtime_error("missing __has_include argument");
+        }
+        simplecpp::Token *tok2 = tok1->next;
```
danmar, line 2355: do we need to handle comments? is this valid code:

    __has_include /*comment*/ ("foo.h")

simplecpp.cpp
```diff
@@ -2341,6 +2341,63 @@ static void simplifySizeof(simplecpp::TokenList &expr, const std::map<std::strin
...
+        if (!tok2) {
+            throw std::runtime_error("missing __has_include argument");
+        }
+        if (tok1->op == '(') {
+            tok1 = tok1->next;
+            while (tok2->op != ')') {
+                tok2 = tok2->next;
```
danmar, line 2362: should we abort if there is an inner "(" ?

simplecpp.cpp
```diff
@@ -2341,6 +2341,63 @@ static void simplifySizeof(simplecpp::TokenList &expr, const std::map<std::strin
     }
 }
 
+/** Evaluate __has_include(file) */
+static std::string openHeader(std::ifstream &f, const simplecpp::DUI &dui, const std::string &sourcefile, const std::string &header, bool systemheader);
+static void simplifyHasInclude(simplecpp::TokenList &expr, const simplecpp::DUI &dui)
+{
+    for (simplecpp::Token *tok = expr.front(); tok; tok = tok->next) {
```
danmar, line 2348: I think it would feel better if there was some testing of this. could we refactor and break up this perhaps so during testing we test if the filename argument for different `__has_include` are extracted properly. i.e. `__has_include("foo.h")` => `"foo.h"`. `__has_include(<path/foo.h>)` => `<path/foo.h>`. ?

danmar: It is my fault there are not tests, but that is making me worried when I review this.

simplecpp.cpp
al42and: According to https://en.cppreference.com/w/cpp/preprocessor/include, having `)` in filenames is legal, so ideally we should check the next token, and:

- if it starts with `"`, just take it as it is, `)` will be a part of it,
- if it starts with `<`, we shall parse till the next `>` token,

and then the next token should be `)`.

But I feel like having `)` in filenames is a pretty obscure edge case.

simplecpp.cpp
al42and: I added several tests here: https://github.com/al42and/simplecpp/commit/c653ae8cf5e4e0f3d20235654b8d6cf1240ac852

Not nearly exhaustive enough, but covers most common cases, IMO.

Please tell me whether you want me to make a PR or whatever.

simplecpp.cpp
al42and: This is valid code, and my testing (https://github.com/al42and/simplecpp/commit/64929e8775dc12011460fbb47795f0b1fea159ea) shows that it's processed correctly.

simplecpp.cpp
ceggers-arri: Hi Andrey,

I haven't found the time to finish this in the previous 6 months and it's unlikely that this will change in short term.

Feel free to take over this branch and finish it.

regards
Christian

danmar:
these changes were merged by pull request 271

PR #248, opened 2022-03-10
added clang-tidy CI job and fixed warnings / added CMake option "DISABLE_CPP03_SYNTAX_CHECK" / CMake cleanups

firewave:
The only interesting suppress warning is currently `clang-analyzer-core.NullDereference` since it depends on very complex value flow, might be a false positive and I didn't have the time to review it.


simplecpp.cpp
```diff
@@ -221,15 +221,17 @@ simplecpp::TokenList::TokenList(std::istream &istr, std::vector<std::string> &fi
     readfile(istr,filename,outputList);
 }
 
-simplecpp::TokenList::TokenList(const TokenList &other) : frontToken(nullptr), backToken(nullptr), files(other.files)
+simplecpp::TokenList::TokenList(const TokenList &other)
```
danmar, line 224: I prefer to copy simplecpp in a separate PR.

firewave:
I disabled `performance-inefficient-string-concatenation` since that warning appears to have false positives - see https://github.com/llvm/llvm-project/issues/54526. I was always curious about this warning but never had the time to properly look into it.

simplecpp.cpp
```diff
@@ -2796,7 +2796,7 @@ static std::string getFileName(const std::map<std::string, simplecpp::TokenList
         return (filedata.find(header) != filedata.end()) ? simplecpp::simplifyPath(header) : "";
     }
 
-    const std::string relativeFilename = getRelativeFileName(sourcefile, header);
```
danmar, line 2799: I just wonder why you remove this const.

danmar (commented):
just a small question

simplecpp.cpp
firewave: That's the `performance-no-automatic-move` check. If a variable is returned and it is declared `const` it cannot be moved but is copied instead.

danmar: so the drawback is less safety. is a const reference a better solution perhaps?

firewave: I will drop this for now.

firewave:
This is ready for review again. I also updated and modernized the clang-tidy job.

simplecpp.cpp
```diff
@@ -1842,7 +1842,7 @@ namespace simplecpp {
                 return recursiveExpandToken(output, temp, loc, tok2, macros, expandedmacros2, parametertokens);
             }
 
-            else if (tok->str() == DEFINED) {
```
danmar, line 1845: please undo all changes in simplecpp.. these should be made in upstream and then they can be copied to cppcheck.

firewave: We are currently in the simplecpp repo. 🙄

danmar: oops!! :-)

PR #249, opened 2022-03-10
moved valgrind to separate build job

firewave:


firewave:
I jumped the gun on this. This actually creates more overhead so I am not sure if this actually makes sense.

PR #251, opened 2022-03-11
added support for built-in defines __DATE__ and __TIME__

firewave:
See https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html.

### DIFF
--- a/test.cpp
+++ b/test.cpp
@@ -2147,6 +2147,47 @@ static void preprocessSizeOf()
     ASSERT_EQUALS("file0,1,syntax_error,failed to evaluate #if condition, invalid sizeof expression\n", toString(outputList));
 }
 
+static void timeDefine()
+{
+    const char code[] = "__TIME__";
+    const std::string t = preprocess(code);
+    // "19:09:53"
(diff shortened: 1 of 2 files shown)


test.cpp
```diff
@@ -2147,6 +2147,47 @@ static void preprocessSizeOf()
...
+
+static void dateDefine()
+{
+    const char code[] = "__DATE__";
+    const std::string dt = preprocess(code);
+    // "\"Mar 11 2022\""
+    ASSERT_EQUALS(13, dt.size());
+    // TODO: split string and check proper ranges instead
+    ASSERT_EQUALS('"', dt[0]);
+    ASSERT_EQUALS(true, dt[1] >= 65 && dt[1] <= 90); // uppercase letter
```
danmar, line 2177: I suggest `dt[1] >= 'A' && dt[1] <= 'Z'`

firewave: D'oh! 🤦‍♂️

firewave: Done.

PR #254, opened 2022-03-15
even more testrunner cleanups

firewave:
This gets rid of all the explicit `std::istringstream` usage in the test cases so I can hopefully easily switch the implementation for the new API introduced with #244.

### DIFF
--- a/test.cpp
+++ b/test.cpp
@@ -86,10 +86,10 @@ static std::string readfile(const char code[], int sz=-1, simplecpp::OutputList 
     return simplecpp::TokenList(istr,files,std::string(),outputList).stringify();
 }
 
-static simplecpp::TokenList makeTokenList(const char code[], std::vector<std::string> &files, const std::string &file)
+static simplecpp::TokenList makeTokenList(const char code[], std::vector<std::string> &filenames, const std::string &filename=std::string(), simplecpp::OutputList *outputList=nullptr)
 {
     std::istringstream istr(code);
-    return simplecpp::TokenList(istr,files,file);
+    return simplecpp::TokenList(istr,filenames,filename,outputList);
 }
 
 static simplecpp::TokenList makeTokenList(const char code[])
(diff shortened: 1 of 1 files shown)


firewave:
@danmar Can this be merged? I need this to finish up that other PR.

test.cpp
```diff
@@ -78,32 +78,41 @@ static void testcase(const std::string &name, void (*f)(), int argc, char * cons
...
+    return simplecpp::TokenList(istr,filenames,filename,outputList);
+}
 
-static std::string readfile(const char code[], int sz=-1, simplecpp::OutputList *outputList=nullptr)
+static simplecpp::TokenList makeTokenList(const char code[])
 {
-    std::istringstream istr(sz == -1 ? std::string(code) : std::string(code,sz));
     std::vector<std::string> files;
-    return simplecpp::TokenList(istr,files,std::string(),outputList).stringify();
+    return makeTokenList(code, files);
```
danmar, line 96: this looks dangerous. the returned TokenList will reference the local `files`

firewave: You are right. I didn't think about this.

We didn't detect this in Cppcheck - I have this enabled in my IDE and it did not report anything.

firewave: I filed https://trac.cppcheck.net/ticket/10975. It is impossible to detect the issue I introduced though since we have no idea what is done with the reference.

firewave: Done.

test.cpp
```diff
@@ -1969,36 +1971,36 @@ static void readfile_string_error()
...
     ASSERT_EQUALS("file0,1,syntax_error,No pair for character (\"). Can't process file. File is either invalid or unicode, which is currently not supported.\n", toString(outputList));
     outputList.clear();
 
-    readfile("A = R\"as\n(abc)as\"", -1, &outputList);
+    readfile("A = R\"as\n(abc)as\"", &outputList);
     ASSERT_EQUALS("file0,1,syntax_error,Invalid newline in raw string delimiter.\n", toString(outputList));
     outputList.clear();
 
-    readfile("A = u8R\"as\n(abc)as\"", -1, &outputList);
+   readfile("A = u8R\"as\n(abc)as\"", &outputList);
```
danmar, line 1986: indentation seems wrong on this line.

firewave: Done.

PR #255, opened 2022-03-15
Special-case ## pasting to string/character constants (issue #168)

patrickdowling:
Building on the approach from PR https://github.com/danmar/simplecpp/pull/169 this special cases the `##` pasting in `expandHashHash` for string and character literals to allow for (user-defined) literals (`operator “”`).

- The string/char case is in it’s own `if` block; it’s not necessarily elegant but trying to maintain a single tokensB/strAB path through the other branches seems harder to follow (\*).
- There’s a dangling TODO because while ## ## or varargs combos appear to work, it’d be great if someone could review that assumption.
- Added some additional `hashhash` test cases. Suggestions for other/better boundary cases welcome.
- In at least two tests (`hashhash18`, `hashhash19`) the result of ## isn’t actually useful (and does eventually produce a syntax error in `cppcheck`). Detecting that seems like a different rabbit hole.

(\*) although now it’s not a trivial merge into cppcheck tree. Can’t win…

(\*\*) As an aside there already exists a difference between the handling of `operator “”` applied to strings/chars or numbers even without macros. But at least with/without macros its now consistent, so this is perhaps a follow-up item.

```
#define STR(a) a ## _str
#define ULL(a) a ## _ull
#define CC(a) a ## _cc

auto s1 = "ABC"_str;
auto s2 = STR("ABC");

auto i1 = 12345_ull;
auto i2 = ULL(12345);

auto c1 = 'A'_cc;
auto c2 = CC('A');
```
```
##AST
[literals.cc:5]
=
|-s1
`-(
  |-operator""_str
  `-,
    |-"ABC" 'const char *'
    `-3 'signed int'

[literals.cc:6]
=
|-s2
`-(
  |-operator""_str
  `-,
    |-"ABC" 'const char *'
    `-3 'signed int'

[literals.cc:8]
=
|-i1
`-12345_ull

[literals.cc:9]
=
|-i2
`-12345_ull

[literals.cc:11]
= 'char'
|-c1 'char'
`-'A' 'char'

[literals.cc:12]
= 'char'
|-c2 'char'
`-'A' 'char'
```

### DIFF
--- a/run-tests.py
+++ b/run-tests.py
@@ -38,7 +38,7 @@ skip = ['assembler-with-cpp.c',
         'has_attribute.cpp',
         'header_lookup1.c', # missing include <stddef.h>
         'line-directive-output.c',
-        'macro_paste_hashhash.c',
+        #  'macro_paste_hashhash.c',
         'microsoft-ext.c',
         'normalize-3.c', # gcc has different output \uAC00 vs \U0000AC00 on cygwin/linux
         'pr63831-1.c', # __has_attribute => works differently on cygwin/linux
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -58,14 +58,16 @@ static bool isOct(const std::string &s)
     return s.size()>1 && (s[0]=='0') && (s[1] >= '0') && (s[1] < '8');
 }
 
-static bool isStringConstant(const std::string &s)
+static bool isStringLiteral(const std::string &s)
 {
     return s.size() > 1 && (s[0]=='\"') && (*s.rbegin()=='\"');
 }
 
-static bool isCharConstant(const std::string &s)
+static bool isCharLiteral(const std::string &s)
 {
-    return s.size() == 3 && (s[0]=='\'') && (s[2]=='\'');
+    // char literal patterns can include 'a', '\t', '\000', '\xff', 'abcd', and maybe ''
+    // This only checks for the surrounding '' but doesn't parse the content.
+    return s.size() > 1 && (s[0]=='\'') && (*s.rbegin()=='\'');
 }
 
 static const simplecpp::TokenString DEFINE("define");
@@ -1931,7 +1933,7 @@ namespace simplecpp {
                 throw invalidHashHash(tok->location, name());
 
             bool canBeConcatenatedWithEqual = A->isOneOf("+-*/%&|^") || A->str() == "<<" || A->str() == ">>";
-            bool canBeConcatenatedStringOrChar = isStringConstant(A->str()) || isCharConstant(A->str());
+            bool canBeConcatenatedStringOrChar = isStringLiteral(A->str()) || isCharLiteral(A->str());
             if (!A->name && !A->number && A->op != ',' && !A->str().empty() && !canBeConcatenatedWithEqual && !canBeConcatenatedStringOrChar)
                 throw invalidHashHash(tok->location, name());
 
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -58,6 +58,15 @@ static bool isOct(const std::string &s)
     return s.size()>1 && (s[0]=='0') && (s[1] >= '0') && (s[1] < '8');
 }
 
+static bool isStringConstant(const std::string &s)
+{
+    return s.size() > 1 && (s[0]=='\"') && (*s.rbegin()=='\"');
+}
+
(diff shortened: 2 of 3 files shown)


simplecpp.cpp
```diff
@@ -58,6 +58,15 @@ static bool isOct(const std::string &s)
...
 }
 
+static bool isStringConstant(const std::string &s)
+{
+    return s.size() > 1 && (s[0]=='\"') && (*s.rbegin()=='\"');
+}
+
+static bool isCharConstant(const std::string &s)
+{
+    return s.size() == 3 && (s[0]=='\'') && (s[2]=='\'');
```
danmar, line 68: imho this should also be classified as a char constant: `'\t'`.
a multi char constant is not handled: `'ab'`. is that intentional?

simplecpp.cpp
```diff
@@ -58,6 +58,15 @@ static bool isOct(const std::string &s)
     return s.size()>1 && (s[0]=='0') && (s[1] >= '0') && (s[1] < '8');
 }
 
+static bool isStringConstant(const std::string &s)
+{
+    return s.size() > 1 && (s[0]=='\"') && (*s.rbegin()=='\"');
```
danmar, line 63: I think this works also: `s.back() == '\"'`

run-tests.py
```diff
@@ -38,7 +38,7 @@ def cleanup(out):
         'has_attribute.cpp',
         'header_lookup1.c', # missing include <stddef.h>
         'line-directive-output.c',
-        'macro_paste_hashhash.c',
+        #  'macro_paste_hashhash.c',
```
danmar, line 41: can you explain what happened here. do you cover everything with your test cases?

simplecpp.cpp
patrickdowling: Ah, good point (and now the comment about '\'' in the original thread makes way more sense).

simplecpp.cpp
patrickdowling: Yeah, that's what I'd prefer but apparently `back()` is c++11 and the compiler complained because the use c++03.

simplecpp.cpp
patrickdowling: multichar constants don't seem to work with operator "" though, so that might be accidentally intentional. Will investigate further.

patrickdowling: So the pragmatic approach seems to be to relax the restriction to `s.size() > 2`. That covers `‘a’`, `‘1234’`, `‘\t’`, `‘\000’` etc. but doesn't check the validity, which seem in line with the regular output. Or even > 1 analog to strings since there’s the weird case of `’’`.

Multi-character plus user-defined operator "" don't seem to actually _compile_ but clang/gcc will happily preprocess it with `-E` ([godbolt](https://godbolt.org/z/GvoExP5Pj)).

simplecpp.cpp
patrickdowling: And by "the" I mean "the test makefile target". So while that's a bit ugly, using `-std=c++03` does indeed give `error: no member named 'back' in 'std::basic_string…` (although admittedly not with every compiler I tried).

simplecpp.cpp
danmar: I didn't know about `''` , yes that is weird :-)

So yes I think the pragmatic approach sounds good.

run-tests.py
danmar: ah sorry I misread as if you removed something.. but if you can comment out this line that is good news!!

simplecpp.cpp
danmar: ok!

danmar:
yay!! I think some users will be very happy!

danmar:
> So, some additional cases to consider down the literals rabbit hole

I assumed this could be investigated in separate PRs. If you want to continue working on that feel free to do it. Otherwise I hope we can open some tickets. The problems you see.. is it simplecpp bugs or cppcheck bugs?

patrickdowling:
Great, thanks for the review!
For those additional u8 cases I haven't actually _seen_ a problem, just after reading up literals I wondered what might happen 😬 . So yeah, that's maybe a separate task. The original ## issues that led me here were actually in cppcheck because it was bailing in a project with `invalidHashHash` before doing anything interesting. That now seems ok - but on that codebase I then have other issues that I haven't investigated deeper yet (but for which I made a trac ticket).

PR #257, opened 2022-03-22
fixed Cppcheck compilation with MSBuild

firewave:
I had to use the Visual Studio IDE to build it to get these issues. They did not happen when using CMake. This could also indicate that the configuration of the CMake Visual Studio build does not match the existing VS project files. Something to look into when we get ready to drop the VS project files.

This removes the `static` specifiers I incorrectly added to some new function in the `simplecpp` name in a58ae6283b9b62ba694034576707c612c3cbce6b. I somehow assumed those are class members.

The issue with duplicated function definitions of `isStringLiteral` and `isCharLiteral` appears to be a compiler bug. There's functions of the same name in `utils.h` in the Cppcheck source. These functions can never see each other but somehow the compiler has a problem with it. See https://youtrack.jetbrains.com/issue/RIDER-64723 for a report on this. This also needs to be reported upstream.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -58,12 +58,14 @@ static bool isOct(const std::string &s)
     return s.size()>1 && (s[0]=='0') && (s[1] >= '0') && (s[1] < '8');
 }
 
-static bool isStringLiteral(const std::string &s)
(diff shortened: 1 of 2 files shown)


danmar:
Thanks! I will try bumping simplecpp again..

PR #261, opened 2022-04-20
test.cpp: also run tests with char buffer

firewave:


test.cpp
```diff
@@ -77,17 +93,39 @@ static void testcase(const std::string &name, void (*f)(), int argc, char * cons
 
 #define TEST_CASE(F)    (testcase(#F, F, argc, argv))
 
+static std::string writeFile(const char code[], std::size_t size, const std::string &filename) {
```
danmar, line 96: I feel that it's a good idea to avoid using real files in tests.

firewave: In general I totally agree with you. But we allow the input of actual files so we should actually test this.

danmar: I still feel skeptic. if we assume that the stream input is bug free then the test should work the same with a string input or a file. if we extrapolate this then this should be done in cppcheck also. I fear I don't like this.

firewave: Sorry for the late reply.

Keeping tests simple is great but I think there should be proper coverage and historically that is something we don't have in either Cppcheck or simplecpp. 😐

Also all the include files are always being read from the disk and not from memory. Not having any tests which rely on temporary files indicates that there is no coverage for reading includes at all. I have not looked into this yet.

So it seems instead of avoiding tests with files on disk we actually need to add much more.

danmar: Having some system test that uses files is good but we do have some such system tests. Testing that APIs work as they should is out of scope.

danmar: > If that is fine with you I will try to finish up that PR without any unit tests.

yes I would prefer that in this case.

firewave: Great.

After thinking about it a bit more the usage with Cppcheck could be considered an integration test.

danmar: yes it's more like a integration test.

firewave:
This could be implemented in the Python tests instead by making this modes available via `simplecpp` (as already is via `-is`). That would require to parameterize each invocation but I am not sure how to do this. @Tal500 any idea?

Tal500:
> available

sorry for the very late reply. see https://github.com/danmar/simplecpp/blob/10c96815e402299da1d3ab1f753dda4c1b0b12f8/integration_test.py#L37-L39

danmar (approved):

PR #266, opened 2022-08-23
upgraded CI to clang-15

firewave:


firewave:
The `misc-const-correctness` false positive is tracked via https://github.com/llvm/llvm-project/issues/57297.

PR #270, opened 2022-08-29
avoid some implicit conversions

firewave:
This does not change any types. It just makes implicit conversions explicit.

Here's the warnings in question for reference:
```
/mnt/s/GitHub/simplecpp-fw/simplecpp.cpp:401:76: warning: implicit conversion loses integer precision: 'int' to 'const unsigned short' [-Wimplicit-int-conversion]
        const unsigned short bom = (static_cast<unsigned char>(istr.get()) << 8);
                             ~~~    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
/mnt/s/GitHub/simplecpp-fw/simplecpp.cpp:541:55: warning: implicit conversion loses integer precision: 'long' to 'unsigned int' [-Wshorten-64-to-32]
                    lineDirective(location.fileIndex, std::atol(numtok->str().c_str()), &location);
                    ~~~~~~~~~~~~~                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/mnt/s/GitHub/simplecpp-fw/simplecpp.cpp:550:35: warning: implicit conversion loses integer precision: 'long' to 'unsigned int' [-Wshorten-64-to-32]
                                  std::atol(numtok->str().c_str()), &location);
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/mnt/s/GitHub/simplecpp-fw/simplecpp.cpp:780:20: warning: implicit conversion loses integer precision: 'int' to 'const char' [-Wimplicit-int-conversion]
    const char c = std::tolower(tok->str()[0]);
               ~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/mnt/s/GitHub/simplecpp-fw/simplecpp.cpp:1266:25: warning: implicit conversion loses integer precision: 'unsigned long' to 'unsigned int' [-Wshorten-64-to-32]
    return files.size() - 1U;
    ~~~~~~ ~~~~~~~~~~~~~^~~~
/mnt/s/GitHub/simplecpp-fw/simplecpp.cpp:2101:35: warning: implicit conversion loses integer precision: 'int' to 'char' [-Wimplicit-int-conversion]
                    windowsPath = toupper(driveLetter);
                                ~ ^~~~~~~~~~~~~~~~~~~~
/mnt/s/GitHub/simplecpp-fw/simplecpp.cpp:2105:35: warning: implicit conversion loses integer precision: 'int' to 'char' [-Wimplicit-int-conversion]
                    windowsPath = toupper(driveLetter);
                                ~ ^~~~~~~~~~~~~~~~~~~~
```

And an example for the warnings with the assertions:
```
/mnt/s/GitHub/simplecpp-fw/test.cpp:1845:29: warning: implicit conversion loses integer precision: 'std::vector::size_type' (aka 'unsigned long') to 'const unsigned int' [-Wshorten-64-to-32]
    ASSERT_EQUALS(1U, files.size());
    ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
```

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -540,7 +540,7 @@ void simplecpp::TokenList::readfile(std::istream &istr, const std::string &filen
                     const Token *numtok = cback();
                     while (numtok->comment)
                         numtok = numtok->previous;
-                    lineDirective(location.fileIndex, std::atol(numtok->str().c_str()), &location);
+                    lineDirective(location.fileIndex, static_cast<unsigned int>(std::atol(numtok->str().c_str())), &location);
                 } else if (lastline == "# %num% %str%" || lastline == "# line %num% %str%") {
                     const Token *strtok = cback();
                     while (strtok->comment)
@@ -549,7 +549,7 @@ void simplecpp::TokenList::readfile(std::istream &istr, const std::string &filen
                     while (numtok->comment)
                         numtok = numtok->previous;
                     lineDirective(fileIndex(replaceAll(strtok->str().substr(1U, strtok->str().size() - 2U),"\\\\","\\")),
-                                  std::atol(numtok->str().c_str()), &location);
+                                  static_cast<unsigned int>(std::atol(numtok->str().c_str())), &location);
                 }
                 // #endfile
                 else if (lastline == "# endfile" && !loc.empty()) {
@@ -1265,7 +1265,7 @@ unsigned int simplecpp::TokenList::fileIndex(const std::string &filename)
             return i;
     }
     files.push_back(filename);
-    return files.size() - 1U;
+    return static_cast<unsigned int>(files.size() - 1U);
 }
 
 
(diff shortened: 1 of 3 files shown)


simplecpp.cpp
```diff
@@ -398,7 +398,7 @@ static unsigned short getAndSkipBOM(std::istream &istr)
 
     // The UTF-16 BOM is 0xfffe or 0xfeff.
     if (ch1 >= 0xfe) {
-        const unsigned short bom = (static_cast<unsigned char>(istr.get()) << 8);
+        const unsigned short bom = static_cast<unsigned short>((static_cast<unsigned char>(istr.get()) << 8));
```
danmar, line 401: I don't see that this change improves the readability or makes the code more robust.

firewave: It makes the implicit conversions explicit so you know what actually happens. If possible the code should be changed to avoid these. Since those might not be straight-forward I did not attempt this yet.

danmar: I just don't agree. The code is longer and I don't see that it makes the code easier. On the contrary..

danmar: The original code is shorter and 100% safe. I consider it a false positive from the compiler to warn about this. There can be no loss of precision.

PR #271, opened 2022-09-17
Improve support for __has_include

al42and:
This PR is, pretty much, #245 with conflict resolution and tests.

The original fix by @ceggers-arri seems perfectly fine: it passes whatever tests I came up with and fixes the problem I had with cppcheck (angle brackets).


danmar (approved):
Looks good to me! There was a CI failure but it seems like something flaky. I will rerun it to see if we can make it green..

PR #285, opened 2023-01-03
some `Token::flags()` improvements

firewave:


test.cpp
```diff
@@ -2424,6 +2424,61 @@ static void cpluscplusDefine()
     ASSERT_EQUALS("\n201103L", preprocess(code, dui));
 }
 
+static void assertToken(const std::string& s, bool name, bool number, bool comment, char op = '\0')
+{
+    const std::vector<std::string> f;
+    const simplecpp::Location l(f);
+    const simplecpp::Token t(s, l);
+    ASSERT_EQUALS(name, t.name);
```
danmar, line 2432: If this ASSERT fails then this line number will be written not where the assertToken is called from. I suggest that we create some helper macros ASSERT_NAME  ASSERT_NUMBER that will report proper line numbers..

firewave: Good catch. I actually ran into this while writing the tests but didn't think about fixing the assert.

PR #286, opened 2023-01-03
improved prerequisite checks for `__has_include` handling

firewave:
The check in `simplecpp::preprocess` obviously didn't work if `gnu++` or another alias were used. Also the simplification should only be performed if supported.

This does not fix the hang tracked in https://trac.cppcheck.net/ticket/11456. The code there is unfortunately a file which is over 1.5 MB big so we need to find a smaller sample.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1980,7 +1980,7 @@ namespace simplecpp {
                             if (tok->next->str() == "(")
                               ++paren;
                             else if (tok->next->str() == ")")
-                              --paren;                            
+                              --paren;
                             if (paren == 0)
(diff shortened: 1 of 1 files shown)


firewave:
Given that we have no default standard set and we don't check the standard in any other place the check in `simplecpp::preprocess()` should probably be removed (as well as the check I additionally added) so we always assume the highest possible standard in all cases.

Or we should assume the highest possible on if it is not specified. That would probably be better.

firewave:
This should be fine to merge. I will address the case of default to the latest standard if none is provided in a different PR and also file a ticket about it.

PR #287, opened 2023-01-10
added `DUI::clearIncludeCache` to clear the non-existing files cache

firewave:
This necessary for the missing include tests in Cppcheck to work. As well for contexts where this exists across multiple scans - this currently lacks proper support in Cppcheck though.


firewave:
This cache is really awkward and undocumented.

I wonder if this should be enabled for all platforms. But that is out-of-scope right now.

PR #302, opened 2023-08-10
clang-tidy.yml: updated to Clang 17

firewave:


firewave:
I think this project is simple enough that we can keep `misc-include-cleaner` enabled.

firewave:
I filed https://github.com/llvm/llvm-project/issues/64587 about the `misc-include-cleaner` false positive.

main.cpp
```diff
@@ -16,11 +16,16 @@
...
+#include <cstring>
 #include <fstream>
 #include <iostream>
-#include <cstring>
+#include <map>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "simplecpp.h"
```
danmar, line 28: It's an advantage to put local headers before the system includes.

the local header might use types in the system includes and then it's good to enforce that there are corresponding includes in the header.

firewave: I accidentally did it the wrong around as `simplecpp.h` is a "system" header in the Cppcheck code base - which this is not...

firewave: Done.

Issue #96, opened 2017-09-01
simplecpp chokes on clang/master/INPUTS/stpcpy-test.c


amai2012:
The following testcode from clang ​https://raw.githubusercontent.com/clang-ykt/clang/master/INPUTS/stpcpy-test.c causes simplecpp to choke (run out of memory):
```c++
#define __extension__

#define __stpcpy(dest, src) (__extension__ (__builtin_constant_p (src) ? (__string2_1bptr_p (src) && strlen (src) + 1 <= 8 ? __stpcpy_small (dest, __stpcpy_args (src), strlen (src) + 1) : ((char *) __mempcpy (dest, src, strlen (src) + 1) - 1)) : __stpcpy (dest, src)))
#define stpcpy(dest, src) __stpcpy (dest, src)
#define __stpcpy_args(src) __extension__ __STRING2_SMALL_GET16 (src, 0), __extension__ __STRING2_SMALL_GET16 (src, 4), __extension__ __STRING2_SMALL_GET32 (src, 0), __extension__ __STRING2_SMALL_GET32 (src, 4)

#define __mempcpy(dest, src, n) (__extension__ (__builtin_constant_p (src) && __builtin_constant_p (n) && __string2_1bptr_p (src) && n <= 8 ? __mempcpy_small (dest, __mempcpy_args (src), n) : __mempcpy (dest, src, n)))
#define mempcpy(dest, src, n) __mempcpy (dest, src, n)
#define __mempcpy_args(src) ((char *) (src))[0], ((char *) (src))[2], ((char *) (src))[4], ((char *) (src))[6], __extension__ __STRING2_SMALL_GET16 (src, 0), __extension__ __STRING2_SMALL_GET16 (src, 4), __extension__ __STRING2_SMALL_GET32 (src, 0), __extension__ __STRING2_SMALL_GET32 (src, 4)

#define __STRING2_SMALL_GET16(src, idx) (((__const unsigned char *) (__const char *) (src))[idx + 1] << 8 | ((__const unsigned char *) (__const char *) (src))[idx])

#define __STRING2_SMALL_GET32(src, idx) (((((__const unsigned char *) (__const char *) (src))[idx + 3] << 8 | ((__const unsigned char *) (__const char *) (src))[idx + 2]) << 8 | ((__const unsigned char *) (__const char *) (src))[idx + 1]) << 8 | ((__const unsigned char *) (__const char *) (src))[idx])

[...]

PR #305, opened 2023-08-21
inverted the relationship between expanded macros and tokens

firewave:
Instead of duplicating the macro references into each token we should just keep references to the tokens in the macros.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1700,6 +1700,11 @@ namespace simplecpp {
                 return invalidHashHash(loc, macroName, "Combining '\\"+ tokenA->str()+ "' and '"+ strAB.substr(tokenA->str().size()) + "' yields universal character '\\" + strAB + "'. This is undefined behavior according to C standard chapter 5.1.1.2, paragraph 4.");
             }
         };
+
+        bool expandsTok(const Token *tok) const
+        {
+            return expandedToks.find(tok) != expandedToks.end();
+        }
     private:
         /** Create new token where Token::macro is set for replaced tokens */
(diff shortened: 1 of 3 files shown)


firewave:
Currently just a hack but tests still pass. And this fixes https://trac.cppcheck.net/ticket/11885.

firewave:
I am pretty sure the hack is all wrong (too warm to think 🫠). But I think the idea is sound 🤞. Definitely needs way more tests first.

firewave:
These changes are better. They pass all the tests in simplecpp and Cppcheck. They do not fix that upstream issue as it hits something different now.

Before
```
Benchmark 1: ./simplecpp -q file.c
  Time (mean ± σ):     433.6 ms ±   4.2 ms    [User: 333.4 ms, System: 99.9 ms]
  Range (min … max):   428.3 ms … 439.6 ms    5 runs
```

After
```
Benchmark 1: ./simplecpp -q file.c
  Time (mean ± σ):     272.8 ms ±   4.8 ms    [User: 223.9 ms, System: 48.5 ms]
  Range (min … max):   266.9 ms … 278.7 ms    5 runs
```

Clang 15 `3,048,566,387` -> `1,328,447,289`

firewave:
Unfortunately these changes case a big regression to the example in #96.

Before
```
Benchmark 1: ./simplecpp -q file2.c
  Time (mean ± σ):      1.088 s ±  0.033 s    [User: 0.784 s, System: 0.304 s]
  Range (min … max):    1.061 s …  1.125 s    5 runs
```

After 
```
Benchmark 1: ./simplecpp -q file2.c
  Time (mean ± σ):      4.029 s ±  0.061 s    [User: 3.292 s, System: 0.734 s]
  Range (min … max):    3.986 s …  4.134 s    5 runs
```

PR #306, opened 2023-08-22
simplecpp.h: push and pop the MSVC warning disabling `#pragma` so they don't spill into user code

firewave:
Those were introduced by #298.


firewave:
I would still prefer not to have warnings disabled in the code but that can still be discussed after. At least fix them for now.

diamante0018 (approved):

PR #312, opened 2023-09-11
refs #352 - generate output error on unknown `DUI::std` value

firewave:
If the value is unknown we are silently setting an invalid/unexcepted value for the predefined macros which might lead to other issues.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -3315,8 +3315,17 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
             macros.insert(std::make_pair("__STDC_VERSION__", Macro("__STDC_VERSION__", std_def, dummy)));
         } else {
             std_def = simplecpp::getCppStdString(dui.std);
-            if (!std_def.empty())
-                macros.insert(std::make_pair("__cplusplus", Macro("__cplusplus", std_def, dummy)));
+            if (std_def.empty()) {
+                if (outputList) {
+                    simplecpp::Output err(files);
+                    err.type = Output::DUI_ERROR;
+                    err.msg = "unknown standard specified: '" + dui.std + "'";
+                    outputList->push_back(err);
+                }
+                output.clear();
+                return;
+            }
+            macros.insert(std::make_pair("__cplusplus", Macro("__cplusplus", std_def, dummy)));
         }
     }
 
--- a/main.cpp
+++ b/main.cpp
@@ -169,6 +169,9 @@ int main(int argc, char **argv)
             case simplecpp::Output::FILE_NOT_FOUND:
                 std::cerr << "file not found: ";
                 break;
+            case simplecpp::Output::DUI_ERROR:
+                std::cerr << "dui error: ";
+                break;
             }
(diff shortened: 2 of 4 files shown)


simplecpp.cpp
```diff
@@ -3219,8 +3219,9 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
             macros.insert(std::make_pair("__STDC_VERSION__", Macro("__STDC_VERSION__", std_def, files)));
         } else {
             std_def = simplecpp::getCppStdString(dui.std);
-            if (!std_def.empty())
-                macros.insert(std::make_pair("__cplusplus", Macro("__cplusplus", std_def, files)));
+            if (std_def.empty())
+                throw std::runtime_error("unknown standard specified: '" + dui.std + "'");
```
danmar, line 3223: I have the feeling that we do not want that this exception is unhandled by simplecpp::preprocess. It would be better to write a error in the output list and return.

danmar: I believe we still want that simplecpp is designed to be portable and easy to reuse.

my gut feeling is that a library should not throw exceptions. if you use compiler A to compile the library and compiler B to compile the executable that uses the library.. is the exceptions "binary" interface well defined enough so A and B will always use the same?

danmar:
I guess a simplecpp issue wouldn't hurt also

firewave:
> I guess a simplecpp issue wouldn't hurt also

Good point. Tickets for behavior changes.

simplecpp.cpp
firewave: I had this sitting in my tree for ages now and I am close to have no pending local changes (finally) so I pushed this. There's no pressing reason for having this. It's just to be verify the given value is actually something valid.

I will take a look if we can handle this differently. Maybe needs some cleanups and definitely some more tests on the Cppcheck side first. The standard handling is questionable in some parts - there's several TODOs in place (or in my local tree).

firewave:
I adjusted the code so it uses the `OutputList` instead of throwing an exception.

Will file a ticket about this later.

firewave:
> Will file a ticket about this later.

Done.

danmar (approved):
lgtm, should I merge?

firewave:
I thought we wanted to make a release with the current state first so we can get this repo and Cppcheck in sync again.

Afterwards feel free to merge anything I made ready for review and you approve of.

danmar:
> I thought we wanted to make a release with the current state first so we can get this repo and Cppcheck in sync again.
> Afterwards feel free to merge anything I made ready for review and you approve of.

Sorry.. I have tagged 1.1.3 now. I tagged the old commit e040047b6bf538fdcdb80d51f0a444d9d0728661

Please feel free to sync cppcheck.

firewave:
@danmar I changed the title of the PR because we do not actually bail out anymore. Strangely your merge still had the old title. I have no idea how this is even possible!?

firewave:
Actually it was merge with the original title from when I created this PR. I changed the title multiple times.

I have no seen this in the Cppcheck repo so far but there I usually merge the PRs myself.

PR #313, opened 2023-09-11
test.cpp: also check the message of the exception

firewave:


firewave:
@danmar anything still to do here?

PR #319, opened 2023-09-12
Fix #314 TokenList::lastLine() is slow for long lines

### LINKED ISSUE #314: TokenList::lastLine() is slow for long lines
`lastLine()` is only called to compare it to short strings, but the result is computed up to 100000 characters. The use of `insert()` causes the runtime to scale as O(N^2) in the number of characters.
https://sourceforge.net/p/cppcheck/discussion/development/thread/4e3ff7c875/

chrchr-github:


chrchr-github:
Thanks for merging this. It appears though that we are still too slow on large files, so we should probably limit maxsize to 1000 tokens.

PR #329, opened 2023-10-23
Added support for `__VA_OPT__` expansion

### LINKED ISSUE #191: C++20 __VA_OPT__
In forum :
https://sourceforge.net/p/cppcheck/discussion/general/thread/2811dc2f3b

It was reported that __VA_OPT__ is not handled.
```
#include <cstdio>
#define p1(fmt, ...)        printf(fmt __VA_OPT__(,) __VA_ARGS__)

void test() {
    p1("hello");
    p1("%s", "hello");
}
```

JohnSiegel:
This change adds support for `__VA_OPT__` in variadic macros. This should close #191

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1959,7 +1959,25 @@ namespace simplecpp {
             // Macro parameter..
             {
                 TokenList temp(files);
-                if (expandArg(&temp, tok, loc, macros, expandedmacros, parametertokens)) {
+                if (tok->str() == "__VA_OPT__") {
+                    if (sameline(tok, tok->next) && tok->next->str() == "(") {
+                        tok = tok->next;
+                        int paren = 1;
+                        while (sameline(tok, tok->next)) {
+                            if (tok->next->str() == "(")
+                              ++paren;
+                            else if (tok->next->str() == ")")
+                              --paren;                            
+                            if (paren == 0)
+                              return tok->next->next;
+                            tok = tok->next;
+                            if (parametertokens.front()->next->str() != ")" && parametertokens.size() > args.size())
+                              tok = expandToken(output, loc, tok, macros, expandedmacros, parametertokens)->previous;
+                        }
+                    }
+                    throw Error(tok->location, "Missing parenthesis for __VA_OPT__(content)");
+                }
+                else if (expandArg(&temp, tok, loc, macros, expandedmacros, parametertokens)) {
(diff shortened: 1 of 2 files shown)


firewave:
Not having check the actual implementation. This should be behind a check if the standard is >= c++20 - possibly similar to what I did in #286.

JohnSiegel:
Here is an example from the unit tests I added:

From `define_va_opt_1`:
```
#define p1(fmt, args...) printf(fmt __VA_OPT__(,) args)

p1(\"hello\");
p1(\"%s\", \"hello\");
```

Output before the change:
```
printf ( "hello" __VA_OPT__ ( , ) ) ;
printf ( "%s" __VA_OPT__ ( , ) "hello" ) ;
```

Output after the change:
```
printf ( "hello" ) ;\n
printf ( "%s" , "hello" ) ;
```

danmar:
@JohnSiegel I think firewaves comment makes sense.

C++11 => this output is correct:

    printf ( "hello" __VA_OPT__ ( , ) ) ;
    printf ( "%s" __VA_OPT__ ( , ) "hello" ) ;

C++20 => this output is correct:

    printf ( "hello" ) ;
    printf ( "%s" , "hello" ) ;

However:
 * gcc seems to handle __VA_OPT__ even if standard is changed to C++11.
 * as far as I see you don't have access to the standard in Macro::expandToken.. so somehow we would need to pass that info.

JohnSiegel:
Sorry, I initially misunderstood @firewave 's comment. Thank you for clarifying @danmar . I can update this PR with a change passing a `simplecpp::DUI&` to `expandToken` and its callers if you think that is the best way to handle this.

danmar:
@JohnSiegel I'd suggest we keep it as it is for now. In general it's a good idea to respect the standard carefully but in this case __VA_OPT__ should be relatively safe to handle. Such identifier should be avoided in older code.

There is a clang-tidy warning. Please fix that.

 /home/runner/work/simplecpp/simplecpp/simplecpp.cpp:1980:17: error: do not use 'else' after 'throw' [readability-else-after-return,-warnings-as-errors]

firewave:
> @JohnSiegel I'd suggest we keep it as it is for now. In general it's a good idea to respect the standard carefully but in this case `__VA_OPT__` should be relatively safe to handle. Such identifier should be avoided in older code.

I can address that in the PR I linked.

PR #330, opened 2023-11-13
mitigated `include-what-you-use` warnings

firewave:
`include-what-you-use` and `clang-include-cleaner` disagree about this. Since we have the latter enabled in this repo appease it for now by re-adding the include and suppress the IWYU warning.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -20,13 +20,13 @@
 #include <cstring>
 #include <ctime>
 #include <exception>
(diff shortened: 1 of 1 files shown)


firewave:
I accidentally modified the file downstream in https://github.com/danmar/cppcheck/commit/f465dc5b3dfb0f7ee409a38bdfe39312ac8fe37e instead of applying them upstream.

PR #333, opened 2024-02-01
Fix #334 Do not assert when source file is missing

### LINKED ISSUE #334: Do not assert when source file is missing
Executing simplecpp with a missing file cause assertion error. We should report some file not found error instead:

```
daniel@laptop:~/simplecpp$ ./simplecpp notexist.cpp 
simplecpp: simplecpp.cpp:388: FileStream::FileStream(const string&): Assertion `file != nullptr' failed. 
Aborted (core dumped)
```

olabetskyi:


simplecpp.cpp
```diff
@@ -385,7 +385,8 @@ class FileStream : public simplecpp::TokenList::Stream {
         , lastCh(0)
         , lastStatus(0)
     {
-        assert(file != nullptr);
+        if (file != nullptr)
```
danmar, line 388: I assume condition should be `file == nullptr`

simplecpp.cpp
```diff
@@ -385,7 +385,8 @@ class FileStream : public simplecpp::TokenList::Stream {
...
+            throw simplecpp::Output::FILE_NOT_FOUND;
```
danmar, line 389: hmm typically throwing a class is prefered. maybe in the short term this enum will work well.. this is not a pattern we want to start using throughout the code I guess.

danmar (commented):
I would think it should be relatively easy to unit test the `simplecpp::TokenList::TokenList` constructor

simplecpp.cpp
```diff
@@ -442,8 +445,15 @@ simplecpp::TokenList::TokenList(std::istream &istr, std::vector<std::string> &fi
...
         : frontToken(nullptr), backToken(nullptr), files(filenames)
 {
-    FileStream stream(filename);
-    readfile(stream,filename,outputList);
+    try
+    {
+        FileStream stream(filename);
+        readfile(stream,filename,outputList);
+    }
+    catch(const simplecpp::Output e) // TODO handle extra type of errors
```
danmar, line 453: please catch by reference.

test.cpp
```diff
@@ -2900,6 +2917,8 @@ int main(int argc, char **argv)
     TEST_CASE(tokenMacro4);
     TEST_CASE(tokenMacro5);
 
+    TEST_CASE(makeTokenListNoFile);
```
danmar, line 2920: hmm.. the test cases are ordered alphabetically more or less.. and I prefer to put related test cases together so somehow I would like to put this near some other test cases where we test the TokenList constructor..

I have the feeling that the "readfile_" tests are somewhat related. They test the raw parsing of a file I think. I suggest you change the name of this test to "readfile_file_not_found"

danmar (approved):
when ci is happy we can merge this

PR #339, opened 2024-02-15
fixed #342/fixed #343 - do not treat directories like regular files in existence checks

### LINKED ISSUE #342: memory leak when processing emoty/non-existent includes with `simplecpp` CLI
```cpp
#include</\\>
#include</\\>
```

```
==58215==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 144 byte(s) in 2 object(s) allocated from:
    #0 0x7ff948ae2002 in operator new(unsigned long) /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_new_delete>
    #1 0x56404715371a in simplecpp::load(simplecpp::TokenList const&, std::vector<std::__cxx11::basic_string<>

Direct leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x7ff948ae2002 in operator new(unsigned long) /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_new_delete>
    #1 0x564047162412 in simplecpp::preprocess(simplecpp::TokenList&, simplecpp::TokenList const&, std::vecto>

SUMMARY: AddressSanitizer: 216 byte(s) leaked in 3 allocation(s).
```

### LINKED ISSUE #343: `#include` pointing to directory is treated as existing empty file
If an include points to an existing directory (e.g. `#include </>`) the code will not raise an error. I would expect it to report `EXPLICIT_INCLUDE_NOT_FOUND`.

This is caused by using `std::ifstream::is_open()` to determine the existence of a file. But if you do this on a directory it will be successful as well as the subsequent read. It will not read anything though so it behaves like an empty file.

firewave:
`std::ifstream` can open directories and read from them without issue. Nothing will be read though - it is like an empty file. This would have treated an include which actually pointed to a directory as existing (e.g. `#include </>`).

### DIFF
--- a/test.cpp
+++ b/test.cpp
@@ -3221,6 +3221,16 @@ static void fuzz_crash()
     }
 }
 
+static void leak()
+{
+    const char code[] = "#include</\\\\>\n"
+                        "#include</\\\\>\n";
+    simplecpp::OutputList outputList;
+    ASSERT_EQUALS("", preprocess(code, &outputList));
+    ASSERT_EQUALS("file0,1,missing_header,Header not found: </\\\\>\n"
+                  "file0,2,missing_header,Header not found: </\\\\>\n", toString(outputList));
+}
+
 int main(int argc, char **argv)
 {
     TEST_CASE(backslash);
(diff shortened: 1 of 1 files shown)


firewave:
No tests added since we still do not have tests for additional files ... but I just realized that certain directories always exists so at least on Linux it can be tested.

Moving code is problematic though as Cppcheck is GPL and this would re-license it under 0BSD.

firewave:
> Moving code is problematic though as Cppcheck is GPL and this would re-license it under 0BSD.

@danmar

After we resolved and merged this we should create another release so we can sync it to Cppcheck.

firewave:
The helper functions need to be adjusted so they align with the current usage in Cppcheck.

PR #340, opened 2024-02-16
refs #342 - do not load included files twice in CLI application / added `DUI::removeComment`

firewave:
The CLI application was pre-loading the included files just to remove the comments. Adding a `DUI` option for that makes it necessary. This makes profiling with the CLI more useful and it also fixes a potential memory leak in the CLI when an empty (or currently also non-existent) include was specified.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -11,7 +11,6 @@
 #include <iostream>
 #include <map>
 #include <string>
-#include <utility>
 #include <vector>
 
 int main(int argc, char **argv)
@@ -110,6 +109,8 @@ int main(int argc, char **argv)
         std::exit(0);
     }
 
+    dui.removeComments = true;
+
     // Perform preprocessing
     simplecpp::OutputList outputList;
     std::vector<std::string> files;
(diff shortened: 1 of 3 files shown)


firewave:
Still a draft as I want to research this a bit more.

Sample which produces a leak (still looking at getting one with valid code):
```cpp
#include  <//.>3=i`/\   >
 88longm a
```

```
==21281==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x7fdc95ee2002 in operator new(unsigned long) /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_new_delete.cpp:95
    #1 0x564f6098d585 in simplecpp::load(simplecpp::TokenList const&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, simplecpp::DUI const&, std::__cxx11::list<simplecpp::Output, std::allocator<simplecpp::Output> >*) /home/user/CLionProjects/simplecpp-rider/simplecpp.cpp:3180

SUMMARY: AddressSanitizer: 72 byte(s) leaked in 1 allocation(s).
```

danmar (approved):
lgtm.. do you feel that this is finished?

firewave:
> lgtm.. do you feel that this is finished?

Yes - on the CLI side this is finished. Otherwise I would have left it a draft :)

PR #347, opened 2024-03-27
made it possible to create a `TokenList` from a buffer

firewave:


firewave:
Still need to integrate/test it with #261 to make sure it does not introduce any issues.

It is already actively used by `Macro` though and no tests are failing, so it should be fine.

firewave:
Some tests are still failing - will investigate.

firewave:
The tests failing are the ones with binary characters in the string. We need to add an additional `const char*, int` input for those.

firewave:
I changed it from a `std::string` input to a `const unsigned char*, std::size_t` and `const char*, std::size_t` (for convienience).

The tests pass but it produces valgrind failures in #261.

firewave:
> The tests pass but it produces valgrind failures in #261.

The issue has been fixed - see https://github.com/danmar/simplecpp/actions/runs/8484373804/job/23247192053?pr=261.

firewave:
I will provide a downstream PR which shows this is working within Cppcheck.

We should also do the (planned) release before merging this.

danmar:
@firewave would you like that I release simplecpp now or do you have some more fixes..?

firewave:
> @firewave would you like that I release simplecpp now or do you have some more fixes..?

Do the release so we are in proper sync again.

Afterwards this PR, #339 and #312 require your feedback - and some minor work from my side.

simplecpp.cpp
```diff
@@ -377,6 +377,42 @@ class StdIStream : public simplecpp::TokenList::Stream {
...
+        : str(str)
+        , size(size)
+        , pos(0)
+        , lastStatus(0)
+    {
+        init();
+    }
+
+    virtual int get() OVERRIDE {
+        if (pos >= size)
```
firewave, line 393: These checks are still a hot spot. No idea if there is a way to implement this in a more performant way.

firewave:
Here is the downstream PR using char buffers instead of streams: https://github.com/danmar/cppcheck/pull/6379.

firewave:
Ready for review since the downstream PR passes all tests.

simplecpp.cpp
```diff
@@ -377,6 +377,42 @@ class StdIStream : public simplecpp::TokenList::Stream {
     std::istream &istr;
 };
 
+class StdCharBufStream : public simplecpp::TokenList::Stream {
```
danmar, line 380: Is this declared in a unnamed namespace?

simplecpp.cpp
```diff
@@ -377,6 +377,42 @@ class StdIStream : public simplecpp::TokenList::Stream {
...
+            return lastStatus = EOF;
+        return str[pos++];
+    }
+    virtual int peek() OVERRIDE {
+        if (pos >= size)
+            return lastStatus = EOF;
+        return str[pos];
+    }
+    virtual void unget() OVERRIDE {
+        --pos;
```
danmar, line 403: spontanous thought.. we don't need to check that pos is greater than zero?

simplecpp.cpp
```diff
@@ -377,6 +377,42 @@ class StdIStream : public simplecpp::TokenList::Stream {
...
+    }
+    virtual bool good() OVERRIDE {
+        return lastStatus != EOF;
+    }
+
+private:
+    const unsigned char *str;
```
danmar, line 410: const unsigned char* const str; ?

firewave: Possibly - will check.

If we want to use that we could enable it in `misc-const-correctness` so it will be used consistently forced across the code.

danmar: then.. feel free to use misc-const-correctness for members. but it's not necessary in this PR.

simplecpp.cpp
danmar: ok maybe we should start allowing c++11 sometime soonish then.

if we want to keep it c++03 compatible we could hide the `namespace {` in a ifdef but I don't know if that is necessary.

But let's look at c++11 in another PR.

danmar (approved):
lgtm

firewave:
> you should have gotten an invite for this project. I hope you can merge..

@danmar Thanks.

simplecpp.cpp
firewave: We already have this enabled so it looks like a false negative.

firewave: The check only checks local variables. I will still file a ticket upstream.

firewave: I forgot to apply this change. Also suggested this upstream: https://github.com/llvm/llvm-project/issues/93282.

PR #350, opened 2024-04-22
do not add empty filename to `files` in `preprocess()`

firewave:


firewave:
Required for https://github.com/danmar/cppcheck/pull/6326.

Not sure if a test can be written for this.

firewave:
Made a draft because the downstream PR which is built upon these changes fails.

firewave:
> Made a draft because the downstream PR which is built upon these changes fails.

Turns out that it was a fluke.

danmar (approved):

firewave:
I will still add a unit test for this.

PR #355, opened 2024-05-23
added Emacs C++ marker to header comment

firewave:


danmar (approved):
lgtm

firewave:
Sorry, I merged it the wrong way. Happened because I didn't have the permissions before and the default differs from what we usually do. Will be fine with the upcoming merges.

PR #356, opened 2024-05-23
fixes #352 - internally default to latest standard if none is provided

### LINKED ISSUE #352: `__cplusplus` not set when `DUI::std` is empty/invalid
The `__cplusplus` macro is not set when the provide `DUI::std` is either empty or invalid. This might lead to unexpected behavior.

firewave:


firewave:
This matches the behavior of Cppcheck.

danmar (approved):

PR #357, opened 2024-07-12
use `<iosfwd>` in header

firewave:


firewave:
The build is stuck in progress because `macos-11` runners are no longer available. That is addressed in #358.

PR #359, opened 2024-07-12
clang-tidy.yml: updated to Clang 19

firewave:


firewave:
Clang 19 release candidate has been released and the apt mirror should also have made the branch switch. As we only utilize the binaries in the CI we can merge this now.

Issue #159, opened 2019-03-14
failed to expand BOOST_MPL_AUX_NA_SPEC, Wrong number of parameters for macro BOOST_PP_TUPLE_ELEM


gdsotirov:
I wasn't sure whether to report this here or in [cppcheck](/danmar/cppcheck) project, so please excuse me if I chose the wrong place.

I stumbled upon the following problem when running cppcheck 1.87 on a C++ source using Boost libraries:

`[/usr/include/boost/preprocessor/arithmetic/sub.hpp:47]: (error) failed to expand 'BOOST_MPL_AUX_NA_SPEC', Wrong number of parameters for macro 'BOOST_PP_TUPLE_ELEM'.`

I checked [this post](https://sourceforge.net/p/cppcheck/discussion/general/thread/8a0a0174/) in the forum, but the error message is not exactly the same. I guess it changed after the fix of #125, but issue #126 is still open and I think #108 is also related to the problem, because the problematic Boost header [include/boost/preprocessor/arithmetic/sub.hpp](https://github.com/boostorg/preprocessor/blob/develop/include/boost/preprocessor/arithmetic/sub.hpp#L47) uses [## operator](http://en.cppreference.com/w/cpp/preprocessor/replace#.23_and_.23.23_operators).

Can I somehow workaround the problem until it's fixed in a new cppcheck version?

danmar:
> I wasn't sure whether to report this here or in cppcheck project, so please excuse me if I chose the wrong place.

I don't know. If the "cppcheck -E" output looks wrong then it should be reported here. We can keep the issue here for now.

> Can I somehow workaround the problem until it's fixed in a new cppcheck version?

Yes certainly.

In one project there is a workaround for boost macro problems. The project has this define:

    #define VARS_TO_NAMES_AND_VARS(...) \
    BOOST_PP_SEQ_FOR_EACH_I(VAR_TO_STRING_COMMA_VAR, % %, BOOST_PP_VARIADIC_TO_SEQ(__VA_ARGS__))

Cppcheck does not handle this code well.

To workaround this the code was updated to:

    #ifndef __cppcheck__
    /// @note Cppcheck does not like the syntax `% %` used here and reports a syntax error if that is seen in analysis.
    /// Maybe someday Cppcheck will have a boost configuration that we can use instead of this ifndef
    #define VARS_TO_NAMES_AND_VARS(...) \
[...]

PR #363, opened 2024-08-24
#pragma once usage to prevent unwanted inclusions from happening even once

datadiode:
This implements the idea behind https://github.com/danmar/simplecpp/issues/159#issuecomment-506986341 in a generic way.
It works by allowing #pragma once to take a wildcard expression to sort out the unwanted include files.

Example from some project of mine, to prevent some inclusions which would otherwise cause cppcheck to abort prematurely:
```C++
#ifdef CPPCHECK
#   pragma once "boost/*"
#   pragma once "google/protobuf/*"
#   pragma once "*.pb.h"
#endif
```
Going with Cppcheck's --include option is particularly convenient for the purpose.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -26,6 +26,7 @@
 #include <limits>
 #include <list>
 #include <map>
+#include <regex>
 #include <set>
 #include <sstream>
 #include <stack>
@@ -3382,6 +3383,15 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
 
     std::set<std::string> pragmaOnce;
 
+    // Support an odd #pragma once usage to prevent unwanted inclusions from happening even once
+    // Example:
+    // #ifdef CPPCHECK
+    // #   pragma once "^boost/"
+    // #   pragma once "^google/protobuf/"
+    // #   pragma once "\.pb\.h$"
+    // #endif
+    std::map<std::string, std::regex> pragmaOddOnce;
+
     includetokenstack.push(rawtokens.cfront());
     for (std::list<std::string>::const_iterator it = dui.includes.begin(); it != dui.includes.end(); ++it) {
         const std::map<std::string, TokenList*>::const_iterator f = filedata.find(*it);
@@ -3508,6 +3518,10 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
 
                 const bool systemheader = (inctok->str()[0] == '<');
                 const std::string header(realFilename(inctok->str().substr(1U, inctok->str().size() - 2U)));
+                if (std::find_if(pragmaOddOnce.begin(), pragmaOddOnce.end(), [&header](auto r) { return std::regex_search(header, r.second); }) != pragmaOddOnce.end()) {
+                    rawtok = gotoNextLine(rawtok);
+                    continue;
+                }
                 std::string header2 = getFileName(filedata, rawtok->location.file(), header, dui, systemheader);
                 if (header2.empty()) {
                     // try to load file..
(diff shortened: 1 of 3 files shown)


firewave:
It is not possible to use C++11 since simplecpp still needs to support C++03.

Also using `std::regex` is highly discouraged because it is so slow that is essentially unusable in any production code.

datadiode:
Would [srell](https://www.akenotsuki.com/misc/srell/en/) be an acceptable replacement for `std::regex`?

firewave:
I guess it is sure to say that any third-party dependency is out of the question.

datadiode:
That seems plausible.
Then I'd rather go for a simplistic wildcard match, preferably by reusing the `fnmatch()` from https://compressionratings.com/d_archiver_template.html.

Tal500:
> Would [SRELL](https://www.akenotsuki.com/misc/srell/en/) be an acceptable replacement for `std::regex`?

@datadiode @firewave FYI: this library is the state of the art solution for compile-time-baking of regexp engine in c++ (source: cppcon first presentation of Herb):
https://github.com/hanickadot/compile-time-regular-expressions

danmar:
> FYI: this library is the state of the art solution for compile-time-baking of regexp engine in c++ 

Thanks for the suggestion.

I think CTRE is awesome!! but it requires modern compilers. and minimum c++17 as far as I know.

simplecpp.cpp
```diff
@@ -3382,6 +3416,15 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
...
     std::set<std::string> pragmaOnce;
 
+    // Support an odd #pragma once usage to prevent unwanted inclusions from happening even once
+    // Example:
+    // #ifdef CPPCHECK
+    // #   pragma once "boost/*"
+    // #   pragma once "google/protobuf/*"
+    // #   pragma once "*.pb.h"
+    // #endif
+    std::set<Mask> pragmaOddOnce;
```
danmar, line 3426: well it could be `std::set<std::string> pragmaOnceMasks;` ..  instead of creating a Mask class what would be the downside to just provide a regular function like:

    bool match(const std::string& mask, const std::string& value, bool i);

danmar:
I don't see that #159 talks about `pragma once` I think it would be better to preprocess the macros properly instead.

I reopened the #159 I think I was too quick to close it before.

I have never seen this `pragma once mask` syntax was it something you invented? I would rather not add this if you invented this syntax.

datadiode:
> I don't see that #159 talks about `pragma once`...

The invented extension to the `pragma once` behavior intends to avoid a need to follow the last-resort idea of hardcoding the exclusion of boost headers like so, as mentioned in said comment:
```
+        if (header.rfind("boost/", 0) == 0)
+            continue;
+
```
#pragma isn't standardized. I don't think it is too wrong to overload it with some esoteric feature which is meaningless to other preprocessors. `pragma ignore_include` may feel less wrong though.

datadiode:
I did think of proposing them a command line option to tell cppcheck to ignore any include paths it picks up from a project file if they match some regular expression. Will go ahead with that idea.

firewave:
> I did think of proposing them a command line option to tell cppcheck to ignore any include paths it picks up from a project file if they match some regular expression. Will go ahead with that idea.

That already exists. It is the `-i` option.

datadiode:
> > I did think of proposing them a command line option to tell cppcheck to ignore any include paths it picks up from a project file if they match some regular expression. Will go ahead with that idea.
> 
> That already exists. It is the `-i` option.

According to the usage information, `-i` _applies only to source files so header files included by source files are not matched_.

firewave:
> According to the usage information, `-i` _applies only to source files so header files included by source files are not matched_.

You are of course right. I might have been thinking of something else (and maybe something I needed at some point).

It might also play into https://trac.cppcheck.net/ticket/12359 and https://trac.cppcheck.net/ticket/13204.

Issue #296, opened 2023-05-09
Wrong handling of #x in macro


danmar:
Example code:

```
#define bar(x) x % 2
#define foo(x) printf(#x "\n")

foo(bar(3));
```

gcc output:

```
printf("bar(3)" "\n");
```

simplecpp output:

```
printf("3%2" "\n");
```

Issue #241, opened 2022-02-21
The standard required whitespace between the preprocessing tokens may be removed


pmor13:
```
$ cat t10.c
#define X(x,y) 	x y
#define STR_(x) #x
#define STR(x) 	STR_(x)
STR(X(Y,Y))

$ ./simplecpp t10.c
"YY"
```
The whitespace is required by the standard.

Extra: running `./simplecpp` built with g++ or clang++ may lead to exit status 127:
```
$ g++ -o simplecpp main.cpp simplecpp.cpp && echo OK && ./simplecpp
OK
# exit status 127

$ g++ --version
g++ (GCC) 11.1.0

$ clang++ -o simplecpp main.cpp simplecpp.cpp && echo OK && ./simplecpp
OK
# exit status 127

$ clang++ --version
clang version 8.0.1 (tags/RELEASE_801/final)

$ LLVM/12.0.0/bin/clang++ -o simplecpp main.cpp simplecpp.cpp && ./simplecpp
Syntax:
simplecpp [options] filename
  -DNAME          Define NAME.
  -IPATH          Include path.
  -include=FILE   Include FILE.
  -UNAME          Undefine NAME.

$ uname -a
CYGWIN_NT-10.0 xxx 3.2.0(0.340/5/3) 2021-03-29 08:42 x86_64 Cygwin
```

Issue #225, opened 2021-08-04
token pasting ## fails with nested concatenation macros


SpareSimian:
Continued from issue #159 

Error message:
testsuite\token_concat.cpp:2: syntax error: failed to expand 'wxSETUPH_PATH', Invalid ## usage when expanding 'wxCONCAT'.

Code to reproduce:
```
#define wxCONCAT(text1, text2)       text1 ## text2
#define wxCONCAT3(x1, x2, x3)       wxCONCAT(wxCONCAT(x1, x2), x3)
#define wxSETUPH_PATH               wxCONCAT3(vc142, _x64, _lib)
extern int wxSETUPH_PATH;
```

gnuless:
This doesn't appear to be valid input.  Where is it from?

The formal arguments to `##` are replaced with their actual arguments before pasting, but they are not further expanded until after pasting.  Thus the expansion of `wxSETUPH_PATH` should proceed as:

```
wxCONCAT3(vc142, _x64, _lib)
wxCONCAT(wxCONCAT(vc142, _x64), _lib)
wxCONCAT(vc143, _x64)  ## _lib
```
The result of `)` pasted with `_lib` isn't a valid token, and so the result is the much-maligned "undefined behavior".  Erroring out is a perfectly acceptable action here.

See ANSI C89 6.8.3.3 or C99 6.10.3.3 et seq.

SpareSimian:
The macros are defined here:

https://github.com/wxWidgets/wxWidgets/blob/master/include/wx/cpp.h

They're used to insert the "#pragma comment" to automatically pull in the correct library (based on compiler version and options) here:

https://github.com/wxWidgets/wxWidgets/blob/master/include/msvc/wx/setup.h

SpareSimian:
Boost does the same thing here:

https://github.com/boostorg/config/blob/develop/include/boost/config/auto_link.hpp

gnuless:
Thanks.  The original `wxCONCAT` from https://github.com/wxWidgets/wxWidgets/blob/master/include/wx/cpp.h reads: 

```
#define wxCONCAT_HELPER(text, line) text ## line
#define wxCONCAT(x1, x2) wxCONCAT_HELPER(x1, x2)
```

Your test case abbreviates this:

```
#define wxCONCAT(text1, text2)       text1 ## text2
```

and that's the source of the problem. The indirection through `wxCONCAT_HELPER` seems useless at first glance, but it's there to satisfy the substitution rules discussed above. 

[...]

Issue #130, opened 2018-05-20
unexpanded inner macro


danmar:
Example code:

    #define MAC2STR(x) x[0],x[1],x[2],x[3],x[4],x[5]
    #define FT_DEBUG(fmt, args...) if(pGlobalCtx && pGlobalCtx->debug_level>=2) printf("FT-dbg: "fmt, ##args)

    FT_DEBUG("  %02x:%02x:%02x:%02x:%02x:%02x\n", MAC2STR(pCtx->wlan_intf_addr[i]));

Simplecpp does not preprocess this code properly. The MAC2STR is not expanded.

PR #364, opened 2024-08-28
Allow reexpansion of currently expanding macros during argument evaluation

### LINKED ISSUE #31: Nested macros are not (always) expanded
See original thread at cppcheck (https://sourceforge.net/p/cppcheck/discussion/general/thread/f6c4e29d/)

Code like this:
```C
#define BUTTON_PORT_NUMBER 2
#define BTN_PIEZO_PIN (0x04)

/// Helper macro to concatenate the parameters
#define CONCATENATOR(x, y) x ## y
/// Helper macro to evaluate the parameters before concatenating
#define EVALUATOR(x, y) CONCATENATOR(x, y)
/// Helper macro to evaluate button port number and add it to 'P'
#define BUTTON_PORT EVALUATOR(P, BUTTON_PORT_NUMBER)

/// Direction port for button
#define BUTTON_PORT_DIR  (EVALUATOR(BUTTON_PORT, DIR))

void func(void)
{
    BUTTON_PORT_DIR &= (uint8_t)(~BTN_PIEZO_PIN);
}
```
Results in this code:
```C
void func ( void )
{
 (  EVALUATOR(P,2)DIR  ) &= ( uint8_t ) ( ~  (  0x04  ) ) ;
}
```
It should be:
```C
void func ( void )
{
 (  P2DIR  ) &= ( uint8_t ) ( ~  (  0x04  ) ) ;
}
```

### LINKED ISSUE #293: Fail to expand macros
Example code:
```
#define bitmask(name) setbits(name##_START, name##_BITS)
#define field(name, value) (((value) << name##_START) & name##_MASK)
#define FOO_START 54
#define FOO_BITS 1
#define FOO_MASK bitmask(FOO)
#define FOO field(FOO, 0b1)
FOO_MASK; 
```
simplecpp output:

a.c:1: syntax error: failed to expand 'FOO_MASK', Invalid ## usage when expanding 'bitmask': Unexpected token ')'

datadiode:
This fixes #31 as reproduced in TEST_CASE(define_define_11) without breaking code from @ydamigos's comment to #225 as reproduced in TEST_CASE(define_define_11a).  It thereby also fixes #293.

Subsequent commits in this PR address #66, #130, #241, #296.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2140,7 +2140,8 @@ namespace simplecpp {
          */
         const Token *expandHash(TokenList *output, const Location &loc, const Token *tok, const MacroMap &macros, const std::set<TokenString> &expandedmacros, const std::vector<const Token*> &parametertokens) const {
             TokenList tokenListHash(files);
-            tok = expandToken(&tokenListHash, loc, tok->next, macros, expandedmacros, parametertokens);
+            const MacroMap macros2; // temporarily bypass macro expansion
+            tok = expandToken(&tokenListHash, loc, tok->next, macros2, expandedmacros, parametertokens);
             std::ostringstream ostr;
             ostr << '\"';
             for (const Token *hashtok = tokenListHash.cfront(); hashtok; hashtok = hashtok->next)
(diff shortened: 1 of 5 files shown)


simplecpp.cpp
```diff
@@ -888,7 +886,7 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
             }
 
             if (prefix.empty())
-                push_back(new Token(s, location)); // push string without newlines
+                push_back(new Token(s, location, isspace(stream.peekChar()))); // push string without newlines
```
danmar, line 889: I believe this is not guaranteed to be portable. we include the `<cctype>` so we should call `std::isspace`.

danmar:
@datadiode I am very sorry for so slow review!! Excellent work.

danmar:
I would just like to replace the isspace with std::isspace but I can try to do that later.

PR #367, opened 2024-09-09
fixed #365 - do not treat standards with empty define string as unknown

### LINKED ISSUE #365: C89/C90 standards (and variants) treated as unknown
```
$ ./simplecpp -std=c89 test.cpp

test.cpp:1: dui error: unknown standard specified: 'c89'
```

The problem is that these standards have no value for `__STDC_VERSION__` so `simplecpp::getCStdString()` returns an empty string which therefore treats it as unknown in `simplecpp::preprocess()`.

firewave:


firewave:
Still a draft while I test it with Cppcheck.

firewave:
Confirmed the fix. Addon tests now pass downstream: https://github.com/danmar/cppcheck/actions/runs/10773958052/job/29874904287?pr=6767.

After this was merged we need a 1.1.5 release so we can downstream this and release a regression fix,

PR #370, opened 2024-09-28
replace 'isspace' with 'std::isspace'

danmar:
I have the feeling that `isspace` is not portable. We include cctype.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -886,7 +886,7 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
             }
 
             if (prefix.empty())
-                push_back(new Token(s, location, isspace(stream.peekChar()))); // push string without newlines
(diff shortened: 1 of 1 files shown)


firewave:
It should be. These C functions exists within and outside the namespace.

danmar:
At least an old C++ Builder did not include ctype.h from cctype but well that was more than 10 years ago. I don't explicitly see that isspace is available in cctype in cppreference i.e. https://en.cppreference.com/w/c/string/byte/isspace.

firewave:
Because you looked at the C standard description. See https://en.cppreference.com/w/cpp/string/byte/isspace for the C++ one.

danmar:
yes as far as I understand that only mentions `std::isspace` .

I compare the text with https://en.cppreference.com/w/cpp/container/vector that does not always provide `std::` explicitly.

PR #382, opened 2024-10-25
fix #381

### LINKED ISSUE #381: When there are header files with the same name, the correct file cannot be found
I created a sample repository to demonstrate this issue, in which there are two header files with the same name `same_name.h`, one of which will not be resolved by SimpleCpp.
This sample code can be correctly compiled and run by Visual Studio. Although this does not cause issues in this example, in my actual scenario, it led to false positives, and the file parsing was interrupted.

Repository path:
https://github.com/foryoung365/SimpleCppHeaderSeachIssue
You can use the following command to reproduce:
```
mkdir build

cppcheck --dump --cppcheck-build-dir=build --output-file=result.xml --xml --project=HeaderSeachIssue.sln --project-configuration="Debug|x64"
```

foryoung365:
```
std::fstream fs(simplePath, std::ios::app);
```
File existence check using fstream with ios::app is implemented as a workaround because:
 - ifstream::good() incorrectly handles directories on Unix-like systems
 - <sys/stat.h> adds platform-specific dependencies
 - It avoids duplication with CppCheck's 'path' implementation
 - std::filesystem (the proper solution) requires C++17 support

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -3125,20 +3125,20 @@ static std::string openHeader(std::ifstream &f, const simplecpp::DUI &dui, const
 
 static bool IsFileExists(const std::string& simplePath)
 {
-    if (simplePath.empty())
-    {
-        return false;
-    }
+	if (simplePath.empty())
+	{
+		return false;
+	}
 
-    // This is a workaround to properly check if a file exists,
+	// This is a workaround to properly check if a file exists,
 	// since std::ifstream::good() incorrectly returns true for directories
-    std::fstream fs(simplePath, std::ios::app);
-    if (fs.good())
-    {
-        return true;
-    }
+	std::fstream fs(simplePath, std::ios::in | std::ios::out);
+	if (fs.good())
+	{
+		return true;
+	}
 
-    return false;
+	return false;
 }
 
 static std::string getFileName(const std::map<std::string, simplecpp::TokenList *> &filedata, const std::string &sourcefile, const std::string &header, const simplecpp::DUI &dui, bool systemheader)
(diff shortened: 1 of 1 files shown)


simplecpp.cpp
```diff
@@ -3123,6 +3123,24 @@ static std::string openHeader(std::ifstream &f, const simplecpp::DUI &dui, const
...
+static bool IsFileExists(const std::string& simplePath)
+{
+	if (simplePath.empty())
+	{
+		return false;
+	}
+
+	// This is a workaround to properly check if a file exists,
+	// since std::ifstream::good() incorrectly returns true for directories
+	std::fstream fs(simplePath, std::ios::app);
```
firewave, line 3135: That is an interesting handling.

firewave:
Thanks. Please give me a few days since I might be out sick.

simplecpp.cpp
foryoung365: This solution still has flaws; when the file does not exist, an empty file will be created. Therefore, I replaced it with
```
std::fstream fs(simplePath, std::ios::in|std::ios::out);
```

Here is a simple test result (linux, mac, and windows):
https://github.com/foryoung365/CppFstreamTest/actions/runs/11549258974

Here are some references:
https://en.cppreference.com/w/cpp/io/basic_filebuf/open

danmar:
how do we test this? you can easily check if a file exists or not in a unit test but if you need to test various edge cases I guess a pytest working in a tempdir is better.

firewave:
The issue referenced here was fixed in #443. So I think this can be closed.

And #339 contains improved code for checking for the existence of a file/directory.

I also filed https://trac.cppcheck.net/ticket/14133 about detecting such code with Cppcheck.

PR #383, opened 2024-11-09
Allow custom pragma processing

tmpvar:
Since pragmas are one of the few extension points of c++, detecting them seems pretty important for various types of tooling.


tmpvar:
this works well for my use-case, but maybe it doesn't make sense for the general lib. I'm going to close this and just float a patch. Feel free to use this if you want.

PR #390, opened 2024-11-22
fixed #387 - do not undefine source-level defines via `-U`

### LINKED ISSUE #387: `-U` incorrectly undefines source-level define
```cpp
#define DEF_1

void f()
{
#ifdef DEF_1
    int i;
#endif
}
```

```
$ ./simplecpp -UDEF_1 test.cpp


void f ( )
{



}
```

firewave:


danmar (approved):

firewave:
Will merge after I prototyped the differentiation of built-in macros.

danmar (approved):

PR #405, opened 2025-01-23
Fix #404: simplecpp::TokenList::constFold does not fold '( 0 ) && 10 < X' properly

### LINKED ISSUE #404: simplecpp::TokenList::constFold does not fold '( 0 ) && 10 < X' properly
Test code:
```
    std::istringstream i("( 0 ) && 10 < X");
    simplecpp::TokenList tokenList(i, filenames);
    tokenList.constFold();
    tokenList.dump();
```
Output:
```
( 0 < X )
```

As `0 && expr` is 0 I believe the output should just be 0.

olabetskyi:
Work in progress

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1296,6 +1296,8 @@ void simplecpp::TokenList::squashTokens(Token *&tok, const std::set<std::string>
                 deleteToken(tok1->*step);
             } else
                 break;
+        } else if (skip) {
+            deleteToken(tok1->*step);
         } else if ((tok1->*step)->op == brackets[0]) {
             ++skip;
             deleteToken(tok1->*step);
@@ -1314,10 +1316,10 @@ static bool checkSideForSingleInt(simplecpp::Token * tok, long long value, bool 
     simplecpp::Token* simplecpp::Token::* const step = forwardDirection ? &simplecpp::Token::next : &simplecpp::Token::previous;
     const char * const brackets = forwardDirection ? "()" : ")(";
     int bracket = 0;
-    bool ret = false;
+    bool found = false;
     for (; tok; tok = tok->*step) {
         if (tok->number && (stringToLL(tok->str()) == value || any)) {
-            ret = true;
+            found = true;
             continue;
         }
         if (tok->op == brackets[0]) {
@@ -1333,7 +1335,7 @@ static bool checkSideForSingleInt(simplecpp::Token * tok, long long value, bool 
         }
         return false;
     }
-    return ret;
+    return found;
 }
 
 static const std::string AND("and");
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1300,9 +1300,9 @@ void simplecpp::TokenList::simpleSquash(Token *&tok, const std::string & result)
 void simplecpp::TokenList::squashTokens(Token *&tok, const std::set<std::string> & breakPoints, bool forwardDirection, const std::string & result)
 {
     const char * const brackets = forwardDirection ? "()" : ")(";
-    Token* Token:: * const step = forwardDirection ? &Token::next : &Token::previous;
+    Token* Token::* const step = forwardDirection ? &Token::next : &Token::previous;
     int skip = 0;
-    Token * tok1 = tok->*step;
+    const Token * tok1 = tok->*step;
     while (tok1 && tok1->*step) {
         if (skip){
             if ((tok1->*step)->op == brackets[1])
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1304,19 +1304,7 @@ static simplecpp::Token * stepBack(simplecpp::Token *tok)
     return tok->previous;
 }
 
-/* simplecpp::Token * simplecpp::TokenList::deleteUntil(Token *tok, const std::string & breakPoint, Token *(step)(Token *))
-{
-    while (tok) {
-        if (tok->str() == breakPoint)
-            return tok;
-        Token * oldToken = tok;
-        tok = (step)(tok);
-        deleteToken(oldToken);
-    }
-    return nullptr;
-} */
-
-int simplecpp::TokenList::getTokensDeleteCount(Token *tok, const std::set<std::string> & breakPoints, Token *(step)(Token *), const std::pair<std::string, std::string> & brackets)
+static int getTokensDeleteCount(simplecpp::Token *tok, const std::set<std::string> & breakPoints, simplecpp::Token *(step)(simplecpp::Token *), const std::pair<std::string, std::string> & brackets)
 {
     int count = 0;
     bool skip = false;
@@ -1340,6 +1328,22 @@ int simplecpp::TokenList::getTokensDeleteCount(Token *tok, const std::set<std::s
     return count;
 }
 
+void simplecpp::TokenList::simpleSquash(Token *tok, const std::string & result)
+{
+    tok = tok->previous;
+    tok->setstr(result);
+    deleteToken(tok->next);
+    deleteToken(tok->next);
+}
+
+void simplecpp::TokenList::squashTokens(Token *tok, const std::set<std::string> & breakPoints, bool forwardDirection, const std::string & result)
+{
+    Token *(*step)(Token *) = forwardDirection ? &stepForward : &stepBack;
+    for (int count = getTokensDeleteCount((step)(tok), breakPoints, step, forwardDirection ? std::make_pair<std::string, std::string>("(", ")") : std::make_pair<std::string, std::string>(")", "(")); count > 0; --count)
+        deleteToken((step)(tok));
+    simpleSquash(tok, result);
+}
+
 static const std::string AND("and");
 static const std::string OR("or");
 void simplecpp::TokenList::constFoldLogicalOp(Token *tok)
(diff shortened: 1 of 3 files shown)


simplecpp.cpp
```diff
@@ -1303,16 +1330,192 @@ void simplecpp::TokenList::constFoldLogicalOp(Token *tok)
...
-        if (!tok->next || !tok->next->number)
+        if (!tok->previous->number && !tok->next->number)
             continue;
 
         int result;
-        if (tok->str() == "||")
-            result = (stringToLL(tok->previous->str()) || stringToLL(tok->next->str()));
-        else /*if (tok->str() == "&&")*/
-            result = (stringToLL(tok->previous->str()) && stringToLL(tok->next->str()));
+        /* std::set<std::string> breakPoints;
```
danmar, line 1339: I think your idea here with breakPoints is good. I am thinking that it should be possible to reuse the same code for "||" and "&&" simplifications.

simplecpp.cpp
```diff
@@ -1303,21 +1347,26 @@ void simplecpp::TokenList::constFoldLogicalOp(Token *tok)
         }
         if (tok->str() != "&&" && tok->str() != "||")
             continue;
-        if (!tok->previous || !tok->previous->number)
+        if (!tok->previous || !tok->next)
             continue;
-        if (!tok->next || !tok->next->number)
+        if (!tok->previous->number && !tok->next->number)
```
danmar, line 1352: I feel it's insufficient to just look at the previous/next token. No folding should occur here:

      X >= 0 || 0 < Y

simplecpp.cpp
```diff
@@ -1282,12 +1270,99 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+}
+
+static simplecpp::Token * stepBack(simplecpp::Token *tok)
+{
+    if (!tok)
+        return nullptr;
+    return tok->previous;
+}
+
+static int getTokensDeleteCount(simplecpp::Token *tok, const std::set<std::string> & breakPoints, simplecpp::Token *(step)(simplecpp::Token *), const std::pair<std::string, std::string> & brackets)
```
danmar, line 1292: hmm I am not sure about the step function pointer . It could be a `bool forward` flag instead. that provides less freedom.

simplecpp.cpp
```diff
@@ -1282,12 +1270,99 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+{
+    int count = 0;
+    bool skip = false;
+    tok = (step)(tok);
```
danmar, line 1296: I would remove the parentheses around `step`

simplecpp.cpp
```diff
@@ -1282,12 +1270,99 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+{
+    if (!tok)
+        return nullptr;
+    return tok->previous;
+}
+
+static int getTokensDeleteCount(simplecpp::Token *tok, const std::set<std::string> & breakPoints, simplecpp::Token *(step)(simplecpp::Token *), const std::pair<std::string, std::string> & brackets)
+{
+    int count = 0;
+    bool skip = false;
```
danmar, line 1295: As far as I see it's not enough with a bool.. there can be inner brackets.

simplecpp.cpp
```diff
@@ -1282,12 +1270,99 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+
+void simplecpp::TokenList::squashTokens(Token *&tok, const std::set<std::string> & breakPoints, bool forwardDirection, const std::string & result)
+{
+    Token *(*step)(Token *) = &stepForward;
+    std::pair<std::string, std::string> brackets = std::make_pair<std::string, std::string>("(", ")");
+    if (!forwardDirection) {
+        step = &stepBack;
+        brackets = std::make_pair<std::string, std::string>(")", "(");
+    }
+    for (int count = getTokensDeleteCount((step)(tok), breakPoints, step, brackets); count > 0; --count)
```
danmar, line 1332: you could just as well delete in the function also as far as I see.

simplecpp.cpp
```diff
@@ -1282,12 +1270,99 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+    tok = tok->previous;
+    tok->setstr(result);
+    deleteToken(tok->next);
+    deleteToken(tok->next);
+}
+
+void simplecpp::TokenList::squashTokens(Token *&tok, const std::set<std::string> & breakPoints, bool forwardDirection, const std::string & result)
+{
+    Token *(*step)(Token *) = &stepForward;
+    std::pair<std::string, std::string> brackets = std::make_pair<std::string, std::string>("(", ")");
```
danmar, line 1327: I am not sure that we should pass `brackets` as a parameter. It's redundant info if we pass direction.

simplecpp.cpp
```diff
@@ -1282,14 +1270,97 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+{
+    tok = tok->previous;
+    tok->setstr(result);
+    deleteToken(tok->next);
+    deleteToken(tok->next);
+}
+
+void simplecpp::TokenList::squashTokens(Token *&tok, const std::set<std::string> & breakPoints, bool forwardDirection, const std::string & result)
+{
+    simplecpp::Token *(*step)(simplecpp::Token *) = &stepForward;
```
danmar, line 1302: It's not really required to create stepForward and stepBack intermediate functions

```suggestion
    Token* (Token:: * const step)() = forwardDirection ? &Token::next : &Token::previous;
```
To use `step` you'd need to use the `->*` operator though: `(tok->*step)()`.

simplecpp.cpp
```diff
@@ -1282,14 +1270,97 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+    std::pair<std::string, std::string> brackets = std::make_pair<std::string, std::string>("(", ")");
```
danmar, line 1303: Or just:
```suggestion
    const char brackets[] = forwardDirection ? "()" : ")(";
```
Instead of `tok2->str() == brackets.first` you can then write `tok2->op == brackets[0]`.
Instead of `tok2->str() == brackets.second` you can then write `tok2->op == brackets[1]`.

simplecpp.cpp
olabetskyi: But those are not functions

simplecpp.cpp
```diff
@@ -1282,14 +1270,88 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+        } else {
+            deleteToken(tok1->*step);
+        }
+    }
+    simpleSquash(tok, result);
+}
+
+static bool checkSideForSingleInt(simplecpp::Token * tok, long long value, bool forwardDirection)
+{
+    simplecpp::Token *(*step)(simplecpp::Token *) = &stepForward;
```
danmar, line 1326: I would prefer to get rid of stepForward and stepBackward.

simplecpp.cpp
```diff
@@ -1282,12 +1270,69 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+
+void simplecpp::TokenList::squashTokens(Token *&tok, const std::set<std::string> & breakPoints, bool forwardDirection, const std::string & result)
+{
+    const char * const brackets = forwardDirection ? "()" : ")(";
+    Token* Token::* const step = forwardDirection ? &Token::next : &Token::previous;
+    int skip = 0;
+    const Token * const tok1 = tok->*step;
+    while (tok1 && tok1->*step) {
+        if (skip){
+            if ((tok1->*step)->op == brackets[1])
```
danmar, line 1294: as far as I see if `((tok1->*step)->op == brackets[1])` is true and skip is 0 , we should stop deleting tokens.

can we get wrong output for instance for  `X && (1 || Y) && Z`

simplecpp.cpp
```diff
@@ -1282,12 +1270,70 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+        if ((tok1->*step)->op == brackets[1]){
+            if (skip) {
+                --skip;
+                deleteToken(tok1->*step);
+            } else
+                break;
+        } else if ((tok1->*step)->op == brackets[0]) {
+            ++skip;
+            deleteToken(tok1->*step);
+        } else if (breakPoints.count((tok1->*step)->str()) != 0) {
```
danmar, line 1302: you should ensure that `skip == 0` here right?

simplecpp.cpp
```diff
@@ -1282,12 +1270,70 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+
+static bool checkSideForSingleInt(simplecpp::Token * tok, long long value, bool forwardDirection, bool any = false)
+{
+    simplecpp::Token* simplecpp::Token::* const step = forwardDirection ? &simplecpp::Token::next : &simplecpp::Token::previous;
+    const char * const brackets = forwardDirection ? "()" : ")(";
+    int bracket = 0;
+    bool ret = false;
+    for (; tok; tok = tok->*step) {
+        if (tok->number && (stringToLL(tok->str()) == value || any)) {
+            ret = true;
```
danmar, line 1320: I feel that you return true immediately rather than setting the variable.

simplecpp.cpp
```diff
@@ -1282,12 +1270,70 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+}
+
+static bool checkSideForSingleInt(simplecpp::Token * tok, long long value, bool forwardDirection, bool any = false)
+{
+    simplecpp::Token* simplecpp::Token::* const step = forwardDirection ? &simplecpp::Token::next : &simplecpp::Token::previous;
+    const char * const brackets = forwardDirection ? "()" : ")(";
+    int bracket = 0;
+    bool ret = false;
+    for (; tok; tok = tok->*step) {
+        if (tok->number && (stringToLL(tok->str()) == value || any)) {
```
danmar, line 1319: I feel this is not enough. If the first token is a number then it will be assumed that the entire side is just an int.
```
X > 0 && Y
```
this function will say that the left hand side of the `&&` is 0.

olabetskyi: It works. I've added a test

simplecpp.cpp
```diff
@@ -1282,14 +1270,74 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+    int skip = 0;
+    const Token * const tok1 = tok->*step;
+    while (tok1 && tok1->*step) {
+        if ((tok1->*step)->op == brackets[1]){
+            if (skip) {
+                --skip;
+                deleteToken(tok1->*step);
+            } else
+                break;
+        } else if (skip) {
```
danmar, line 1299: so if skip is true here then we don't count inner brackets that seems unfortunate

simplecpp.cpp
```diff
@@ -1282,12 +1270,52 @@ void simplecpp::TokenList::constFoldBitwise(Token *tok)
...
+                deleteToken(tok1->*step);
+            } else
+                break;
+        } else if ((tok1->*step)->op == brackets[0]) {
+            ++skip;
+            deleteToken(tok1->*step);
+        } else if (skip) {
+            deleteToken(tok1->*step);
+        } else if (breakPoints.count((tok1->*step)->str()) != 0) {
+            deleteToken(tok1->*step);
```
danmar, line 1305: It does not seem obvious to me that this token should be deleted if it's a breakpoint?

olabetskyi: It was a leftover. I removed it and added test to cover this case

danmar (approved):
If CI is happy we can merge this

PR #421, opened 2025-03-17
clang-tidy.yml: updated to Clang 21

firewave:


glankk:
> @glankk this was introduced in [8f8bfe3](https://github.com/danmar/simplecpp/commit/8f8bfe37186b372e93d907c5059a19c730d07b7e) - please have a look

The root cause of this issue seems to be that `Macro` doesn't have a move constructor, so its copy constructor is used instead in the `std::pair` move constructor. The default move constructor in `Macro` is deleted because it has a reference as a member (which I think is generally a bad idea).

I think the cleanest solution to this is to replace the references with pointers and implement sane move operators. I'll work out a draft. @danmar please chime in if you have an opinion on this.

glankk:
Related: #493

firewave:
I filed https://trac.cppcheck.net/ticket/14119 about detecting one of the `misc-const-correctness` cases ourselves.

firewave:
I worked around the warning for now and filed #537.

simplecpp.h
```diff
@@ -430,6 +430,7 @@ namespace simplecpp {
         std::pair<FileData *, bool> get(const std::string &sourcefile, const std::string &header, const DUI &dui, bool systemheader, std::vector<std::string> &filenames, OutputList *outputList);
 
         void insert(FileData data) {
+            // NOLINTNEXTLINE(misc-const-correctness) - FP
             FileData *const newdata = new FileData(std::move(data));
```
firewave, line 434: I filed https://github.com/llvm/llvm-project/issues/157730 about this.

simplecpp.cpp
```diff
@@ -1927,7 +1932,7 @@ namespace simplecpp {
                 }
             }
 
-            Token * const output_end_1 = output.back();
+            const Token * const output_end_1 = output.back();
```
chrchr-github, line 1935: This is not semantically const. Maybe a NOLINT would be more appropriate.

firewave: Done.

simplecpp.cpp
```diff
@@ -2232,7 +2237,7 @@ namespace simplecpp {
             const bool canBeConcatenatedStringOrChar = isStringLiteral_(A->str()) || isCharLiteral_(A->str());
             const bool unexpectedA = (!A->name && !A->number && !A->str().empty() && !canBeConcatenatedWithEqual && !canBeConcatenatedStringOrChar);
 
-            Token * const B = tok->next->next;
+            const Token * const B = tok->next->next;
```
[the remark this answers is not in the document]
firewave: There's several false negatives.

simplecpp.cpp
```diff
@@ -2545,7 +2551,7 @@ static void simplifySizeof(simplecpp::TokenList &expr, const std::map<std::strin
         }
 
         std::string type;
-        for (simplecpp::Token *typeToken = tok1; typeToken != tok2; typeToken = typeToken->next) {
+        for (const simplecpp::Token *typeToken = tok1; typeToken != tok2; typeToken = typeToken->next) {
```
firewave, line 2554: False negative tracked as https://trac.cppcheck.net/ticket/14119.

simplecpp.cpp
```diff
@@ -2526,11 +2532,11 @@ static void simplifySizeof(simplecpp::TokenList &expr, const std::map<std::strin
     for (simplecpp::Token *tok = expr.front(); tok; tok = tok->next) {
         if (tok->str() != "sizeof")
             continue;
-        simplecpp::Token *tok1 = tok->next;
+        const simplecpp::Token *tok1 = tok->next;
```
firewave, line 2535: I filed https://trac.cppcheck.net/ticket/14161 about this.

simplecpp.cpp
```diff
@@ -1329,7 +1329,7 @@ void simplecpp::TokenList::constFoldLogicalOp(Token *tok)
 void simplecpp::TokenList::constFoldQuestionOp(Token **tok1)
 {
     bool gotoTok1 = false;
-    for (Token *tok = *tok1; tok && tok->op != ')'; tok =  gotoTok1 ? *tok1 : tok->next) {
+    for (const Token *tok = *tok1; tok && tok->op != ')'; tok =  gotoTok1 ? *tok1 : tok->next) {
```
firewave, line 1332: This is also only technically correct.

firewave: Done.

firewave:
I validated the `misc-const-correctness` changes with #548 and the suppressions removed.

simplecpp.cpp
```diff
@@ -1329,6 +1329,7 @@ void simplecpp::TokenList::constFoldLogicalOp(Token *tok)
 void simplecpp::TokenList::constFoldQuestionOp(Token **tok1)
 {
     bool gotoTok1 = false;
+    // NOLINTNEXTLINE(misc-const-correctness) - technically correct but used to access non-cost data
```
danmar, line 1332: "non-cost" => "non-const"

simplecpp.cpp
```diff
@@ -1945,6 +1951,7 @@ namespace simplecpp {
                 }
             }
 
+            // NOLINTNEXTLINE(misc-const-correctness) - technically correct but used to access non-cost data
```
danmar, line 1954: spelling: "non-cost"

danmar (approved):
I just have spelling to complain about

simplecpp.cpp
firewave: Done.

simplecpp.cpp
firewave: Done.

PR #429, opened 2025-05-21
fix: parent relative paths, and rework on the whole path extraction mechanics

Tal500:
- Include also another integration test function to check header in parent dir usage (that was failing in previous commit).
- Add a safety check that we did not meet the header file that we are about to insert to the map (abort with stderr message otherwise) - this not only detect unmet bugs in including paths, but also avoid memory leak.
- Add much more cover in the test functions, and check more details.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2725,30 +2725,28 @@ static std::string dirPath(const std::string& path) {
     const std::size_t firstSlash = path.find_last_of("\\/");
     if (firstSlash == std::string::npos) {
         return "";
-    } else {
-        return path.substr(0, firstSlash + (withTrailingSlash ? 1U : 0U));
     }
+    return path.substr(0, firstSlash + (withTrailingSlash ? 1U : 0U));
 }
 
 static std::string omitPathTrailingSlash(const std::string& path) {
     if (endsWith(path, "/")) {
         return path.substr(0, path.size() - 1U);
-    } else {
-        return path;
     }
+    return path;
 }
 
 static std::string extractRelativePathFromAbsolute(const std::string& absoluteSimplifiedPath, const std::string& prefixSimplifiedAbsoluteDir = currentDirectory()) {
     const std::string normalizedAbsolutePath = omitPathTrailingSlash(absoluteSimplifiedPath);
     std::string currentPrefix = omitPathTrailingSlash(prefixSimplifiedAbsoluteDir);
-    std::string leadingParenting = "";
+    std::string leadingParenting;
     while (!startsWith_(normalizedAbsolutePath, currentPrefix)) {
         leadingParenting = "../" + leadingParenting;
         currentPrefix = dirPath<false>(currentPrefix);
     }
     const std::size_t size = currentPrefix.size();
     std::string relativeFromMeetingPath = normalizedAbsolutePath.substr(size, normalizedAbsolutePath.size() - size);
-    if (startsWith_(relativeFromMeetingPath.c_str(), "/")) {
+    if (startsWith_(relativeFromMeetingPath, "/")) {
         // omit the leading slash
         relativeFromMeetingPath = relativeFromMeetingPath.substr(1, relativeFromMeetingPath.size());
     }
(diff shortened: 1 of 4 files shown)


simplecpp.cpp
```diff
@@ -2719,14 +2719,38 @@ static std::string toAbsolutePath(const std::string& path) {
     return simplecpp::simplifyPath(path);
 }
 
-static std::pair<std::string, bool> extractRelativePathFromAbsolute(const std::string& absolutepath) {
-    static const std::string prefix = currentDirectory() + "/";
-    if (startsWith_(absolutepath, prefix)) {
-        const std::size_t size = prefix.size();
-        return std::make_pair(absolutepath.substr(size, absolutepath.size() - size), true);
+// templated function for compiler optimization opportunities
```
danmar, line 2722: I hesitate to use templates unless the reason is to make code generic and avoid copy paste.

simplecpp.cpp
```diff
@@ -2719,14 +2719,38 @@ static std::string toAbsolutePath(const std::string& path) {
...
+template <bool withTrailingSlash>
+static std::string dirPath(const std::string& path) {
+    const std::size_t firstSlash = path.find_last_of("\\/");
```
danmar, line 2725: rename firstSlash to lastSlash

danmar (commented):
some comments. I didn't check it all carefully yet.

simplecpp.cpp
Tal500: OK, there will be no need for that on the final version. I'm working on fixing some crucial errors, and have more detailed tests, so see the next version.

simplecpp.cpp
Tal500: fixed

simplecpp.cpp
```diff
@@ -2719,14 +2719,36 @@ static std::string toAbsolutePath(const std::string& path) {
...
+        return path.substr(0, path.size() - 1U);
+    }
+    return path;
+}
+
+static std::string extractRelativePathFromAbsolute(const std::string& absoluteSimplifiedPath, const std::string& prefixSimplifiedAbsoluteDir = currentDirectory()) {
+    const std::string normalizedAbsolutePath = omitPathTrailingSlash(absoluteSimplifiedPath);
+    std::string currentPrefix = omitPathTrailingSlash(prefixSimplifiedAbsoluteDir);
+    std::string leadingParenting;
+    while (!startsWith_(normalizedAbsolutePath, currentPrefix)) {
```
danmar, line 2741: how does this work when a relative path can't be created.. i.e. if current directory is "C:\test1" then what is the relative path for "D:\foo"

danmar (commented):
small question. partial review.

simplecpp.cpp
danmar: > What it should do in this case is to return ../foo. Is it working? 

that will not work in windows. `../foo` would point at `C:\foo` if current folder is `C:\test1`.
And something like `../../d:/foo` will not work neither.

Tal500: > > What it should do in this case is to return ../foo. Is it working?
> 
> that will not work in windows. `../foo` would point at `C:\foo` if current folder is `C:\test1`. And something like `../../d:/foo` will not work neither.

Oh I missed the `D` here. Anyway, this case will never happen - since in the first place if the `#include` directive is relative it will never can break outside of the partition `C:`.

danmar: Ideally we would have a regression test for 6ee95f7 however I'm not sure how to do that. A pytest would require that there was different drives right? or can we test that the absolute path is returned even if it does not exist?

danmar: for your information, feel free to ping me if it takes couple of days for a response.

simplecpp.cpp
```diff
@@ -2719,14 +2719,42 @@ static std::string toAbsolutePath(const std::string& path) {
...
+    const std::size_t lastSlash = path.find_last_of("\\/");
+    if (lastSlash == std::string::npos) {
+        return "";
     }
-    // otherwise
-    return std::make_pair("", false);
+    return path.substr(0, lastSlash + (withTrailingSlash ? 1U : 0U));
+}
+
+static std::string omitPathTrailingSlash(const std::string& path) {
```
danmar, line 2730: hmm.. Ideally we would unit test such small utility functions. But I also like that they are static. I don't know can it be solved in an elegant way.

PR #430, opened 2025-05-21
fix: leak + double header tokenizing in simplecpp::load

Tal500:
This was catched using the leak sanitizer that is part of the address sanitizer.

On the line `ret[header2] = tokens;`, sometimes there is already a key `header2` in the `ret` map, causing both overriding(leak) and wasteful header tokenizing (two times at least).


Tal500:
Closing as unneeded, since the memory leak was found and avoided correctly in #429

Issue #337, opened 2024-02-13
Line splicing in comment not handled properly


chrchr-github:
https://trac.cppcheck.net/ticket/12379
~~~
#include <string>
#include <vector>
#include <iostream>

void f()
{
        std::size_t i;
        const std::vector<std::string> v{"4", "2"};
        // Some comment\
        for(i = 0U; i < v.size(); ++i)
        {
                std::cout << v[i];
        }
}
~~~

PR #431, opened 2025-05-23
fix #12379: FN: uninitvar due to line splicing

clock999:


clock999:
Here are the corresponding links about this issue:
https://sourceforge.net/p/cppcheck/discussion/general/thread/37c6eba151/
https://github.com/danmar/simplecpp/issues/337

simplecpp.cpp
```diff
@@ -788,6 +788,35 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
             while (stream.good() && ch != '\r' && ch != '\n') {
                 currentToken += ch;
                 ch = stream.readChar();
+                if(ch == '\\') {
```
danmar, line 791: can't you just use `peakChar` instead, something like:
```
    if (ch == '\\' && std::strchr("\r\n", stream.peakChar())) {
        // continue comment on next line..
        currentToken += "\\" + stream.readChar();
        continue;
    }
```
I don't know if `good()` should be used here.
please check if we handle windows "\r\n" line endings also.

danmar: or a flag..
```
     bool backslash = false;
     while (stream.good() && ((ch != '\r' && ch != '\n') || backslash) {
         currentToken += ch;
         backslash = (ch == '\\');
         ch = stream.readChar();
     }
```

danmar:
The PR title is misleading. The fix is to handle line splicing differently. simplecpp doesn't contain any uninitvar checking and there is no uninitialized variable usage I hope.

There should be a proper issue in simplecpp issue tracker.

simplecpp.cpp
```diff
@@ -788,6 +788,35 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
             while (stream.good() && ch != '\r' && ch != '\n') {
```
danmar, line 788: I am thinking if something like this could work:
```
    char lastChar = ' ';
    while (stream.good() && (lastChar == '\\' || (ch != '\r' && ch != '\n')) {
        currentToken += ch;
        lastChar = ch;
        ch = stream.readChar();
    }
```

test.cpp
```diff
@@ -434,7 +434,7 @@ static void comment_multiline()
     const char code[] = "#define ABC {// \\\n"
                         "}\n"
                         "void f() ABC\n";
-    ASSERT_EQUALS("\n\nvoid f ( ) { }", preprocess(code));
+    ASSERT_EQUALS("\n\nvoid f ( ) {", preprocess(code));
```
danmar, line 437: is this the only line splicing test we have?

I think we should test what happens when there are multiple newlines after the \:
```
//  x \\\n
\n
\n
...
```
and an explicit test for \r\n:
```
//  x \\\r\n
...
```
and I wonder what the behavior (according to standard / according to gcc+msvc+etc) is if there is space after the backslash:
```
// x \\ \n
...
```

simplecpp.cpp
```diff
@@ -788,6 +788,35 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
...
                 ch = stream.readChar();
+                if(ch == '\\') {
+                    TokenString tmp;
+                    char tmp_ch = ch;
+                    while((stream.good()) && (tmp_ch == '\\')) {
+                        tmp += tmp_ch;
+                        tmp_ch = stream.readChar();
+                    }
+                    if(!stream.good()) {
+                        if(!tmp.empty()) {
```
danmar, line 799: this `if (!tmp.empty())` is redundant.

simplecpp.cpp
```diff
@@ -788,6 +788,35 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
...
+                            currentToken += tmp_ch;
+                            tmp_ch = stream.readChar();
+                        }
+                        if(!stream.good()) {
+                            break;
+                        }
+                    }
+
+                    ch = tmp_ch;
+                    continue;
```
danmar, line 818: this continue is not need right?

simplecpp.cpp
```diff
@@ -788,6 +788,35 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
...
+
+                    if(tmp_ch != '\r' && tmp_ch != '\n') {
+                        currentToken += tmp;
+                    } else {
+                        while((stream.good()) && (tmp_ch == '\r' || tmp_ch == '\n')) {
+                            currentToken += tmp_ch;
+                            tmp_ch = stream.readChar();
+                        }
+                        if(!stream.good()) {
+                            break;
```
danmar, line 813: this break can be removed right since the outer loop will break if stream is not good.

test.cpp
danmar: I don't remember exactly how the location is adjusted right now. Will the locations below the line splicing comment be correct? i.e.:
```
// hello \
world
x=1;
```
will `x=1;` line number be 3?

clock999: > is this the only line splicing test we have?
> 
> I think we should test what happens when there are multiple newlines after the :
> 
> ```
> //  x \\\n
> \n
> \n
> ...
> ```
> 
> and an explicit test for \r\n:
> 
> ```
> //  x \\\r\n
> ...
> ```
> 
> and I wonder what the behavior (according to standard / according to gcc+msvc+etc) is if there is space after the backslash:
> 
> ```
> // x \\ \n
> ...
> ```
Hi Danmar,
```
I wonder what the behavior (according to standard / according to gcc+msvc+etc) is if there is space after the backslash
```
I checked this with the gcc. The gcc version is 13.3.0. 
If there is space after the backslash, the next line will be commented. That is different with the case there is a non-space char after the backslash.
```
#include <stdio.h>

int main() {
    int x;
    //abc \    
    printf("there is a space\n");
    //abc \x
    printf("there is a char x\n");
    //abc \
    printf("there is nothing\n");
    return 0;
}

Output:
there is a char x
```
The 1st and 3rd output both didn't happen.
Interetingly, the editor seems have a different behavior with the gcc. I used the geditor on ubuntu. The editor will display the comment line with a color of purple. But it doesn't treat the first print as a comment line.
comment
But obviously I think we should follow the rule of the gcc.

clock999: I am checking the unit test case and corresponding code.
```
static void backslash()
{
...
    readfile("//123 \\ \n456", &outputList);
    ASSERT_EQUALS("file0,1,portability_backslash,Combination 'backslash space newline' is not portable.\n", toString(outputList));
...
}
```

```
            const std::string::size_type pos = currentToken.find_last_not_of(" \t");
            if (pos < currentToken.size() - 1U && currentToken[pos] == '\\')
                portabilityBackslash(outputList, files, location);
            if (currentToken[currentToken.size() - 1U] == '\\') {
                ++multiline;
                currentToken.erase(currentToken.size() - 1U);
            } else {
                stream.ungetChar();
            }
```

clock999: > I don't remember exactly how the location is adjusted right now. Will the locations below the line splicing comment be correct? i.e.:
> 
> ```
> // hello \
> world
> x=1;
> ```
> 
> will `x=1;` line number be 3?

Hi Danmar, I checked the code with --debug set on.
```
int main() {
    //abc \
    123
    return 0;
}
```
The debug output:
```
##file /home/w/Desktop/tmp/123.c
1: int main ( ) {
2:
3:
4: return 0 ;
5: }

##Value flow
File /home/w01/Desktop/tmp/4.c
Line 4
  0 always 0

```
The line number should be correctly handled.

simplecpp.cpp
clock999: Hi Danmar,
I see there is a check in the unit tests:
```
readfile("//123 \\ \n456", &outputList);
ASSERT_EQUALS("file0,1,portability_backslash,Combination 'backslash space newline' is not portable.\n", 
toString(outputList));
```
Seems that means if there is a space or tab('\t') after a backslash, such a warning needs to be added to the outputList. Do you think that is necessary? Sorry for troubling you with this question. I don't know why we need to do that. I updated the commit, but just removed the above test case.

danmar: I do have a feeling that it's a portability issue that we would like to warn about. I have the feeling that the standard only talks about `<backslash><newline>` not `<backslash><whitespace><newline>`. If you can see that the standard define the behavior for `<backslash><whitespace><newline>` then a warning is not needed.

test.cpp
```diff
@@ -176,8 +176,8 @@ static void backslash()
 
     readfile("//123 \\\n456", &outputList);
     ASSERT_EQUALS("", toString(outputList));
-    readfile("//123 \\ \n456", &outputList);
-    ASSERT_EQUALS("file0,1,portability_backslash,Combination 'backslash space newline' is not portable.\n", toString(outputList));
+    //readfile("//123 \\ \n456", &outputList);
```
danmar, line 179: I just want that a portability warning is written here. uncomment this test. Then I think we can merge this.

clock999: Hi Danmar,
I uncommented the test, and added the warning to pass the unit test.

PR #433, opened 2025-06-03
Keep comments

osesov:
Keep comments in the main text, not only in defines


danmar:
Cppcheck needs to be able to fetch comments. But off the top of my head it fetches the comments from the "raw tokenlists". cppcheck does not expect comment tokens in the "preprocess" output if I am not mistaken.

osesov:
```
#define XY(X,Y)    X /*a*/ ## /*b*/ Y
```
you right these expansion do not work. No reason to apply this change then.

PR #435, opened 2025-06-11
main.cpp: error out when file/path provided by `-I` or `-include=` do not exist

firewave:


firewave:
It will only fail when `-is` is specified. The check within simplecpp is still flawed.

These checks are important for #438. The code can be simplified if #339 has been merged.

firewave:
> It will only fail when `-is` is specified. The check within simplecpp is still flawed.

I added checks for all cases in the CLI. I filed #539 about improving the library.

PR #436, opened 2025-06-11
main.cpp: added option `-f` to fail on output error

firewave:


danmar (commented):
I don't have strong opinions as I see it this is more or less a "demo" or "test" tool.

But `-f` will often mean `--force` as far as I know.

How about the option `-E` , `-W` or `-Werror` instead.

firewave:
> I don't have strong opinions as I see it this is more or less a "demo" or "test" tool.

Yes, it is not a drop-in for something else so we can make up any options as we go.

danmar (approved):

PR #437, opened 2025-06-11
selfcheck.sh: actually fail if we encountered unexpected errors

firewave:


firewave:
Requires #436 to be merged first.

danmar (approved):

PR #438, opened 2025-06-11
selfcheck.sh: also run with system includes made available

firewave:


firewave:
It now fails with

```
/usr/include/c++/v1/__availability:78: syntax error: failed to evaluate #if condition, undefined function-like macro invocation: __has_feature( ... )
```

or 

```
/usr/include/c++/v1/__assert:29: syntax error: failed to evaluate #if condition, undefined function-like macro invocation: __has_builtin( ... )
```

danmar (commented):
hmm.. my problem with this could be that maybe there will be problem in CI and no problem locally .. or no problem in CI today and then we get problems a year from now.

are the headers in a github runner fixed so they are never changed? or can they be changed at anytime (from one day to the next)?

firewave:
> Yes, that is a bit of a shortcoming. That might need some adjustment if encountered.

This can be improved upon later by getting the built-in include paths from the compiler (if possible).

selfcheck.sh
```diff
@@ -1,11 +1,111 @@
-#!/bin/sh
+#!/bin/bash
```
firewave, line 1: Needed so the `<<<` syntax works.

danmar (approved):

PR #441, opened 2025-06-12
Fix #391 (`__TIME__` replacement might be empty depending on compiler)

### LINKED ISSUE #391: `__TIME__` replacement might be empty depending on compiler
The `%T` format specifier we are using to generate the timestamp is not supported by all compilers:

```
/home/runner/work/simplecpp/simplecpp/simplecpp.cpp: In function ‘std::string getTimeDefine(const tm*)’:
/home/runner/work/simplecpp/simplecpp/simplecpp.cpp:3291:34: error: ISO C++98 does not support the ‘%T’ gnu_strftime format [-Werror=format=]
 3291 |     strftime(buf, sizeof(buf), "%T", timep);
      |                                  ^
```

glankk:
The `%T` format specifier for `strftime` is somehow still not implemented in mingw-w64, causing simplecpp to expand `__TIME__` to an empty string. This is an issue with the compiler suite, but the easiest fix is to use the equivalent `%H:%M:%S` instead. C99's `%T` is defined to be equivalent to `%H:%M:%S`, so this change should be  compatible with all implementations supporting `strftime`.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -3440,7 +3440,7 @@ static std::string getDateDefine(const struct tm *timep)
 static std::string getTimeDefine(const struct tm *timep)
 {
     char buf[] = "??:??:??";
-    strftime(buf, sizeof(buf), "%T", timep);
+    strftime(buf, sizeof(buf), "%H:%M:%S", timep);
     return std::string("\"").append(buf).append("\"");
 }
 


danmar:
please create an issue.
if a test can be added that would be great.
I guess you could install mingw-w64 with this: https://github.com/marketplace/actions/install-mingw

firewave:
Actually #475 confirmed that this fails and requires this change to succeed. So merging it.

PR #444, opened 2025-06-15
change the rules of relativeness preserving to depend on the source file including it for relative path includes

Tal500:
The specification from https://github.com/danmar/cppcheck/pull/7554#issuecomment-2933900683 :

A path (both for (1) top-level source files and (2) header files - i.e. files that are introduced by `#include`) will be absolute if and only if **at least** one of the condition is true:
1. The top-level source file (for headers it is the one that includes it) is given as an input file to simplecpp in its absolute form, **except** for the case that this is a header file that its inclusion was resolved by inclusion dirs (i.e. `-I`) - and this condition is false in that latter case (note that sometimes condition 3 can be true in that case, resulting in absolute flavor).
2. For header files only that are not system includes(i.e. `#include "..."`): The include path is absolute.
3. For header files that were resolved by inclusion dirs (i.e. `-I`): The resolved inclusion dir is absolute.

Whenever the same file is introduced twice or more, only the first occurrence that simplecpp sees will determine the absoluteness flavor.

### DIFF
--- a/integration_test.py
+++ b/integration_test.py
@@ -66,7 +66,7 @@ def test_relative_header_2(record_property, tmpdir, with_pragma_once, inv, sourc
     record_property("stderr", stderr)
     if with_pragma_once:
         assert stderr == ''
-        if inv:
+        if inv or not source_relative:
             assert f'#line 8 "{pathlib.PurePath(tmpdir).as_posix()}/test.h"' in stdout
         else:
             assert '#line 8 "test.h"' in stdout
@@ -93,16 +93,17 @@ def test_relative_header_3(record_property, tmpdir, is_sys, inv, source_relative
         assert "missing header: Header not found" in stderr
     else:
         assert stderr == ''
-        if inv:
-            assert f'#line 8 "{pathlib.PurePath(test_subdir).as_posix()}/test.h"' in stdout
-        else:
+        if source_relative and not inv:
             assert '#line 8 "test_subdir/test.h"' in stdout
+        else:
+            assert f'#line 8 "{pathlib.PurePath(test_subdir).as_posix()}/test.h"' in stdout
 
 @pytest.mark.parametrize("use_short_path", (False, True))
 @pytest.mark.parametrize("relative_include_dir", (False, True))
 @pytest.mark.parametrize("is_sys", (False, True))
 @pytest.mark.parametrize("inv", (False, True))
-def test_relative_header_4(record_property, tmpdir, use_short_path, relative_include_dir, is_sys, inv):
+@pytest.mark.parametrize("source_relative", (False, True))
+def test_relative_header_4(record_property, tmpdir, use_short_path, relative_include_dir, is_sys, inv, source_relative):
     test_subdir = os.path.join(tmpdir, "test_subdir")
     os.mkdir(test_subdir)
     header_file, _ = __test_relative_header_create_header(test_subdir)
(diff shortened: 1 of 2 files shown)


danmar:
Thanks! Let's try this..

danmar:
@Tal500 This worked fine. However it did not compile when I merged to master :-(
```
daniel@laptop:~/simplecpp$ make -j12
g++ -Wall -Wextra -pedantic -Wcast-qual -Wfloat-equal -Wmissing-declarations -Wmissing-format-attribute -Wredundant-decls -Wundef -Wno-multichar -Wold-style-cast -std=c++0x -g -c simplecpp.cpp
g++ -Wall -Wextra -pedantic -Wcast-qual -Wfloat-equal -Wmissing-declarations -Wmissing-format-attribute -Wredundant-decls -Wundef -Wno-multichar -Wold-style-cast -std=c++0x -g -c main.cpp
simplecpp.cpp: In function ‘std::string getFileIdPath(const std::map<std::__cxx11::basic_string<char>, simplecpp::TokenList*>&, const std::string&, const std::string&, const simplecpp::DUI&, bool)’:
simplecpp.cpp:3283:23: error: ‘relativeOrAbsoluteFilename’ was not declared in this scope
 3283 |         openHeader(f, relativeOrAbsoluteFilename);
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

Tal500:
> @Tal500 I am trying to fix it.

The fix is clear - please rename the variable of the @glankk commit from the "old" `relativeOrAbsoluteFilename` to the updated `absoluteFilename` var that this PR does, and the code shall compile (this PR intentionally converted the intermediate result to be always absolute).

Tal500:
> @Tal500 I will let you take a proper look at this. As far as I see the new test does fail without the glankk fix.
> 
> Also.. we have performance issues in simplecpp in windows. and glankk has been able to identify that the `realFileName` is causing it. We discussed yesterday and glankk suggest to use some "file id" instead of "file name" to distinguish files. I think glankk can explain it better in next week than I can.. but I would appreciate feedback from you about this.

My obvious guess is that the performance bottleneck is because of the lock (MyLock). To test if my suggestion is correct, you can omit the lock and measure the performance on a single threaded execution.

As far as I can see, you have two options to reduce the intense usage of a lock:
1. (the simple one) Use the "file exist cache" per thread (i.e. executor, it is the DUI parameter if I understand correctly?). If two threads check with the OS if the file exists, nothing happen.
2. A more complex mechanism to handle atomical (unordered possibly) map, maybe with per thread reflected cache. But I believe this way is complicated and we wouldn't get meaningful performance gain from that, comparing to 1.

BTW: My guess is that you can improve performance if you'd use `std::unordered_map` in the condition that the user cpp version is high enough (C++ 11 right?)

BTW2: Isn't there any API in windows to check if a file exist? It might be faster than actual opening the file.

I do not understand how file "id" will give better performance than file "name"/"path".

danmar:
> According to this [blog](https://thamara.dev/posts/pragma-once-vs-header-guards/), symbolic link is consider to be a different file in pragma once for both msvc and gcc.

ok. My guess is that this has not been an active decision then. The blog says it's a caveat not a feature. 

> In general I think it would make the behavior more correct and uniform across platforms.

This is the feeling I have actually.

> Maybe more importantly though, it allows #pragma once to work correctly with symlinks and case insensitive systems mounted on linux, which matches the behavior of gcc and clang. In general I think it would make the behavior more correct and uniform across platforms.

I am basically interested to see if file ids will provide a speedup and simplify the realFileName code. Therefore I suggest that @glankk makes some proof of concept. I do hope you @Tal500 will review and see if there are problems..

> My obvious guess is that the performance bottleneck is because of the lock (MyLock). To test if my suggestion is correct, you can omit the lock and measure the performance on a single threaded execution.

This is something we can also test.

Tal500:
> > According to this [blog](https://thamara.dev/posts/pragma-once-vs-header-guards/), symbolic link is consider to be a different file in pragma once for both msvc and gcc.
> 
> ok. My guess is that this has not been an active decision then. The blog says it's a caveat not a feature.

I shall say that according to cppreference, #pragma once is implementation defined behaviour, meaning we need to test the output of compilers in Windows on symlinks, and case insensitive paths, assuming we want to mimic the majority compilers behaviour.

PR #445, opened 2025-06-20
Fix #446 (change the rules of relativeness preserving to depend on the source file including it for relative path includes)

### LINKED ISSUE #446: change the rules of relativeness preserving to depend on the source file including it for relative path includes
cppcheck tests was failing because relative paths was converted to absolute paths too much in warning messages.

The specification from https://github.com/danmar/cppcheck/pull/7554#issuecomment-2933900683 :

A path (both for (1) top-level source files and (2) header files - i.e. files that are introduced by #include) will be absolute if and only if at least one of the condition is true:

 *   The top-level source file (for headers it is the one that includes it) is given as an input file to simplecpp in its absolute form, except for the case that this is a header file that its inclusion was resolved by inclusion dirs (i.e. -I) - and this condition is false in that latter case (note that sometimes condition 3 can be true in that case, resulting in absolute flavor).
 *   For header files only that are not system includes(i.e. #include "..."): The include path is absolute.
 *   For header files that were resolved by inclusion dirs (i.e. -I): The resolved inclusion dir is absolute.

Whenever the same file is introduced twice or more, only the first occurrence that simplecpp sees will determine the absoluteness flavor.

danmar:
fixup, #444

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -3173,7 +3173,7 @@ static std::string openHeader(std::ifstream &f, const std::string &path)
     return "";
 }
 
-static std::string getRelativeFileName(const std::string &baseFile, const std::string &header)
+static std::string getRelativeFileName(const std::string &baseFile, const std::string &header, bool returnAbsolutePath)
 {
     const std::string baseFileSimplified = simplecpp::simplifyPath(baseFile);
     const std::string baseFileAbsolute = isAbsolutePath(baseFileSimplified) ?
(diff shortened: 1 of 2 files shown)


simplecpp.cpp
```diff
@@ -3273,14 +3273,14 @@ static std::string getFileIdPath(const std::map<std::string, simplecpp::TokenLis
     }
 
     if (!systemheader) {
-        const std::string relativeOrAbsoluteFilename = getRelativeFileName(sourcefile, header);// unknown if absolute or relative, but always simplified
-        const std::string match = findPathInMapBothRelativeAndAbsolute(filedata, relativeOrAbsoluteFilename);
+        const std::string relativeFilename = getRelativeFileName(sourcefile, header, true);
```
danmar, line 3276: @Tal500 I changed this variable name to relativeFilename.. is that ok?

Tal500: Sounds great!

PR #447, opened 2025-06-24
Include and path handling optimization

glankk:
This is a draft with some suggested changes to how includes are handled. There are several style issues at the moment, but I would like to discuss the overall approach before addressing those.

The changes to the include lookup procedure are based on the observation that files are very often included multiple times, but very rarely included with a different name. To take advantage of this, the lookup procedure first tries to do an optimistic match using the quick and simple path canonicalization provided by `simplifyPath`. In the rare case that a file is included with a name that hasn't been seen before (due to absoluteness change, capitalization, or link), then and only then is that file compared against the cached files using a more accurate but expensive procedure.

For the case where the file identity cannot be determined with `simplifyPath`, the file id is used (`FILE_ID_INFO` on windows, `st_dev` & `st_ino` on linux). This combined with the fact that file identity checking is delayed until absolutely necessary means that all calls to `realFilename` have been removed. This is where most of the timesave comes from.

Using file id means that the lookup procedure has less platform-specific code (only for getting and comparing the file id), and the behavior for case-sensitivty and links is uniform across platforms. The behavior for symlinks matches that of gcc and clang, which is to allows pragma once to work when a file is included with a different name. There's no standard definition for what pragma once should do, but personally I think this is the desired behavior. Intuitively, there are no situations where it is desirable for `#pragma once` to produce redefinition errors if a file is included multiple times using different names.

The code runs about 2 minutes faster for a large test case on windows with many nested includes (12m40s down from 14m40s), though it's a contrived test case so I'm not sure how representative that is of a real use case.

Some other things;
- The new procedure requires some additional metadata. Instead of adding more static data in simplecpp, I've opted to place both the token lists and metadata in a new class for the sake of better structure. Other file-specific optimizations could also be facilitated by this such as the multiple-include optimization used by gcc, though using a class is not strictly necessary and does incur a change to the simplecpp interface.
- Some c++11 features are used, but these could be removed or placed in ifdefs.
- The rules for how simplecpp should search for headers are not always clear, ideally this would be stated precisely somewhere to make sure that the implementation is correct. Specifically, gcc and clang does not look in the current working directory unless `-I.` is explicitly passed, but simplecpp does. I would like to clarify whether or not this is intentional or if the behavior should match other compilers.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -28,7 +28,6 @@
 #include <limits>
 #include <list>
 #include <map>
-#include <memory>
 #include <set>
 #include <sstream>
 #include <stack>
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -2988,18 +2988,18 @@ static std::string openHeader(std::ifstream &f, const simplecpp::DUI &dui, const
     return "";
 }
 
-std::pair<simplecpp::FileData *, bool> simplecpp::FileDataCache::tryload(const FileDataCache::name_map_type::iterator &name_it, const simplecpp::DUI &dui, std::vector<std::string> &filenames, simplecpp::OutputList *outputList)
+std::pair<simplecpp::FileData *, bool> simplecpp::FileDataCache::tryload(FileDataCache::name_map_type::iterator &name_it, const simplecpp::DUI &dui, std::vector<std::string> &filenames, simplecpp::OutputList *outputList)
 {
     const std::string &path = name_it->first;
     FileID fileId;
 
     if (!getFileId(path, fileId))
-        return std::make_pair(nullptr, false);
+        return {nullptr, false};
 
     const auto id_it = mIdMap.find(fileId);
     if (id_it != mIdMap.end()) {
         name_it->second = id_it->second;
-        return std::make_pair(id_it->second, false);
+        return {id_it->second, false};
     }
 
     std::ifstream f(path);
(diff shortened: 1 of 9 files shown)


danmar:
I don't mind at all that realFilename is removed. So imho feel free to continue. I would like to see how the tests will be affected..

glankk:
I've made some additional changes to the path handling, including a pretty big simplification of how path strings are computed. This means that many helper functions have been able to be removed, which I think makes the code cleaner and easier to follow, and also a bit faster.
Tests are passing in simplecpp and cppcheck. I've made a draft of the cppcheck integration here: https://github.com/danmar/cppcheck/pull/7635

simplecpp.cpp
```diff
@@ -3126,206 +2928,182 @@ class NonExistingFilesCache {
...
+    for (const auto &includePath : dui.includePaths) {
+        std::string path = openHeaderDirect(f, simplecpp::simplifyPath(includePath + "/" + header));
         if (!path.empty())
             return path;
     }
     return "";
 }
 
-static std::string openHeader(std::ifstream &f, const simplecpp::DUI &dui, const std::string &sourcefile, const std::string &header, bool systemheader)
+std::pair<simplecpp::FileData *, bool> simplecpp::FileDataCache::tryload(const FileDataCache::name_map_type::iterator &name_it, const simplecpp::DUI &dui, std::vector<std::string> &filenames, simplecpp::OutputList *outputList)
```
danmar, line 2991: I think that the interface would be a bit more intuitive if `name_it` is non-const because the `name_it->second` is assigned in the function.

simplecpp.cpp
```diff
@@ -3126,206 +2928,182 @@ class NonExistingFilesCache {
...
-    // otherwise - try to use the normalize to the correct representation
-    std::string alternativePath;
-    if (isAbsolutePath(path)) {
-        alternativePath = extractRelativePathFromAbsolute(simplecpp::simplifyPath(path));
-    } else {
-        alternativePath = toAbsolutePath(path);
-    }
+    name_it->second = data;
+    mIdMap.insert(std::make_pair(fileId, data));
+    mData.push_back(std::unique_ptr<FileData>(data));
```
danmar, line 3013: Now that we have C++11 I think we can cleanup:
```suggestion
    mIdMap.emplace(fileId, data);
    mData.emplace_back(data);
```

simplecpp.h
```diff
@@ -381,6 +394,119 @@ namespace simplecpp {
...
+        FileDataCache &operator=(FileDataCache &&) = default;
+
+        /** Get the cached data for a file, or load and then return it if it isn't cached.
+         *  returns the file data and true if the file was loaded, false if it was cached. */
+        std::pair<FileData *, bool> get(const std::string &sourcefile, const std::string &header, const DUI &dui, bool systemheader, std::vector<std::string> &filenames, OutputList *outputList);
+
+        void insert(FileData data) {
+            FileData *const newdata = new FileData(std::move(data));
+
+            mData.push_back(std::unique_ptr<FileData>(newdata));
```
danmar, line 422: use either make_unique or emplace. as far as I see a make_unique would mean we can remove the `new` so I guess I prefer that..

danmar (commented):
nits.
I am pretty happy about this and feel that the code is a lot cleaner overall.

simplecpp.h
glankk: make_unique is c++14, so I'll use emplace

simplecpp.cpp
glankk: Alright

PR #448, opened 2025-06-26
ENH: Add support to find Headers in Apple Frameworks

hjmjohnson:
The header locations in apple frameworks require manipulation of the directory name in a '<pkg/myhdr.h>' include defintion.

The header file is located at <pkg.framework/Headers/myhdr.h> in a framework.

### DIFF
--- a/test.cpp
+++ b/test.cpp
@@ -2084,6 +2084,32 @@ static void systemInclude()
     ASSERT_EQUALS("", toString(outputList));
 }
 
+#if 0 // Disabled until the test can be written properly
+#ifdef  __APPLE__
+static void appleFrameworkIncludeTest()
+{
+    // This test is for the Apple framework include handling.
+    // If -I /Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/System/Library/Frameworks
+    // then:
+    const char code[] = "#include <Foundation/Foundation.h>\n";
+    // should find the include file:
+    // /Library/Developer/CommandLineTools/SDKs/MacOSX14.5.sdk/System/Library/Frameworks/Foundation.framework/Headers/Foundation.h
+    std::vector<std::string> files;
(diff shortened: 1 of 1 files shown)


danmar:
Thanks for your contribution.

An issue would be nice.

And some test.

We haven't switched to C++11 yet so that this library will be as portable as possible. However I think we can start supporting C++11. I think it should be widely supported by now. :-)

danmar:
> It might be a few weeks before I circle back to push changes upstream.

ok thanks understood.. for your information we have a PR to start using C++11 so if you get back in few weeks I will hopefully not complain about you using C++11 at least.

hjmjohnson:
@danmar I have updated this PR, but I am struggling to figure out how to make a test to exercise this code.  I tried to decipher how other tests were working, but could not identify how the "cache" would need to be setup to test the new functionality.  Any guidance would be greatly appreciated.

Thanks,
Hans

glankk:
I suggest we emulate gcc's darwin-specific `-F` option, and possibly `-iframework` as well. See https://gcc.gnu.org/onlinedocs/gcc/Darwin-Options.html. The `-F` paths are interleaved with `-I` and searched left-to-right, so the order of interleaved normal include paths and frameworks should be preserved. For this we could change the `DUI::includePaths` from being a vector of path strings to being a vector of structs with a path string and, on darwin, a framework flag that determines what search rules to use.

danmar:
I  will go on vacation for a week. Please ping me later if you want review..

hjmjohnson:
@glankk This is likely a great suggestion, but it is increasing the scope of effort beyond what I can contribute in the short term.  

Do you have the bandwidth to make the changes that you recommend, if so, please take whatever is useful from my suggestion and incorporate it.

firewave:
The tests would probably be best implemented in Python.

PR #450, opened 2025-06-30
Fix #449 (Update c++ standard to c++11)

### LINKED ISSUE #449: Update c++ standard to c++11

glankk:


diamante0018:
No, I don't have any objections

firewave:
@danmar Well, that's up to you to decide. I am not aware of switching to C++11 helping with any of the major known problems, so it's an arbitrary decision.

If the ticket or the PR had some reasoning there might be something to base a discussion on.

danmar:
> If the ticket or the PR had some reasoning there might be something to base a discussion on.

It's just that C++11 has some nicer syntax and more robustness in general.. it is not a specific problem.

firewave:
> It's just that C++11 has some nicer syntax and more robustness in general.. it is not a specific problem.

The question is how many people have this integrated into a C++03 codebase but I guess that feedback will come after the change.

It would allow to fix some minor warnings from various tooling. But taking the previous paragraph in mind we should hold off on those for a few weeks.

firewave:
Probably makes sense to enable the two C++11 warnings for Clang in `CMakeLists.txt` as well.

datadiode:
Thanks for asking. No objections.

hjmjohnson:
FYI: Clang-tidy has some great code analysis tools, supporting C++11 and later.  I am looking forward to this transition!

PR #451, opened 2025-06-30
Fix #368 (__VA_OPT__ is not handled good enough)

### LINKED ISSUE #368: __VA_OPT__ is not handled good enough
Example code:

    #define P( x, ...) printf( x __VA_OPT__(,) __VA_ARGS__)
    #define PF( x, ...) P( x __VA_OPT__(,) __VA_ARGS__)

    int main()
    {
        PF( "%s", "Hello" );
    }

The output from simplecpp is:

    int main ( )
    {
     printf  ( "%s"  __VA_OPT__  (  ,  ) "Hello"  ) ;
    }

glankk:
Fixes `__VA_OPT__` by expanding it first, before further recursive expansions in variadic macros.

When a variadic macro containing a `__VA_OPT__` token is parsed, the macro is split into a different token list for the two possible cases, where each occurence of `__VA_OPT__` is either expanded and discarded. The correct token list is selected during macro expansion by examining the number of arguments, this way most of the extra processing required is only done once during the macro parsing, and only for variadic macros containing `__VA_OPT__`.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1956,7 +1956,7 @@ namespace simplecpp {
             const Token *endToken2;
 
             if (variadicOpt) {
-                if (parametertokens2.size() > args.size() && parametertokens2.front()->next->op != ')')
+                if (parametertokens2.size() > args.size() && parametertokens2[args.size() - 1]->next->op != ')')
                     valueToken2 = optExpandValue->cfront();
                 else
                     valueToken2 = optNoExpandValue->cfront();
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1764,7 +1764,7 @@ namespace simplecpp {
                 for (const Token *tok = valueToken; tok && tok != endToken;) {
                     if (tok->str() == "__VA_OPT__") {
                         if (!sameline(tok, tok->next) || tok->next->op != '(')
-                            throw Error(tok->location, "Missing opening parenthesis for __VA_OPT__");
+                            throw Error(tok->location, "In definition of '" + nameTokDef->str() + "': Missing opening parenthesis for __VA_OPT__");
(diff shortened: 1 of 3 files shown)


simplecpp.cpp
```diff
@@ -2338,6 +2388,13 @@ namespace simplecpp {
         /** is macro variadic? */
         bool variadic;
 
+        /** does the macro expansion have __VA_OPT__? */
+        bool variadicOpt;
+
+        /** Expansion value for varadic macros with __VA_OPT__ expanded and discarded respecively */
```
danmar, line 2394: missing "t": respecively / respectively

test.cpp
```diff
@@ -934,7 +934,7 @@ static void define_va_opt_3()
                          "err()";
 
     ASSERT_EQUALS("", preprocess(code2, &outputList));
-    ASSERT_EQUALS("file0,1,syntax_error,failed to expand 'err', Missing parenthesis for __VA_OPT__(content)\n",
+    ASSERT_EQUALS("file0,1,syntax_error,Failed to parse #define, Missing opening parenthesis for __VA_OPT__\n",
```
danmar, line 937: hmm.. I personally have a better feeling for "expand" than "parse" is this reported for unexpanded macros also? 

Imho it seems helpful to say the macro name.

I like that you say "opening" / "closing" :+1:

simplecpp.cpp
glankk: Fixed

test.cpp
glankk: The error happens on the `#define` directive during parsing and not during macro expansion, so the error is emitted even if the macro is never expanded. This is the same behavior as GCC and Clang. I've update the message to include the macro name.

PR #453, opened 2025-07-01
Fix #452: Undefined function-style macro does not cause an error

### LINKED ISSUE #452: Undefined function-style macro does not cause an error
Example input:
```c
#if A(1)
#endif
```

Undefined macros should only be expanded to 0 if they are non-functional or used without parentheses.

ludviggunne:


simplecpp.cpp
```diff
@@ -2832,9 +2832,20 @@ static void simplifyName(simplecpp::TokenList &expr)
                 if (alt)
                     continue;
             }
+            if (tok->next && tok->next->str() == "(") {
+                if (outputList) {
```
danmar, line 2836: to me it looks like you can more or less use `throw std::runtime_error("Undefined function-like macro in directive: " + tok->str() + "( ... )")` here and the proper syntax error will be written in the output list. but the location will not be correct then will it? would it be better to throw an other error?

danmar (approved):

PR #456, opened 2025-07-02
Fix #454: Accept __has_include for GNU C standards

### LINKED ISSUE #454: Accept `__has_include` for GNU C standards.
Using `__has_include` with `-std=gnu...` currently fails for C standards and C++ standards < 17:

`file.c`:
```c
#if __has_include(<dir/file.h>)
#endif
```

Output:
```console
$ ./simplecpp -std=gnu99 -e file.c
file.c:1: syntax error: failed to evaluate #if condition, division/modulo by zero
```

ludviggunne:


simplecpp.h
```diff
@@ -337,12 +337,13 @@ namespace simplecpp {
...
      */
     struct SIMPLECPP_LIB DUI {
-        DUI() : clearIncludeCache(false), removeComments(false) {}
+        DUI() : gnu(false), clearIncludeCache(false), removeComments(false) {}
         std::list<std::string> defines;
         std::set<std::string> undefined;
         std::list<std::string> includePaths;
         std::list<std::string> includes;
         std::string std;
+        bool gnu;
```
danmar, line 346: not a very strong opinion but if `gnu` is a method instead then std and gnu can't mismatch. i.e.:
```
    bool gnu() const {
         return std.rfind("gnu",0) == 0;
    }
```

ludviggunne: cppcheck removes the "gnu" when creating the DUI:
```c++
    // TODO: use mSettings.standards.stdValue instead
    if (lang == Standards::Language::CPP) {
        dui.std = mSettings.standards.getCPP();
    ...
```
I'm not sure why `mSettings.standards.stdValue` is not used and is only a TODO here, but currently I think we need to have the gnu boolean, since cppcheck would turn for example `gnu++11` in to `c++11`.

firewave: > I'm not sure why `mSettings.standards.stdValue` is not used and is only a TODO here, but currently I think we need to have the gnu boolean, since cppcheck would turn for example `gnu++11` in to `c++11`.

Because I do not want to change stuff without having a test case and try to make dedicated fixes instead of drive-by ones - so I keep dropping TODOs in the code. And I keep getting distracted by too much other stuff...

ludviggunne: > simplecpp should set `__STRICT_ANSI__=1` if a non-`gnu` standard is provided.

Should I create an issue?

ludviggunne: https://github.com/danmar/simplecpp/issues/459

danmar (approved):

PR #460, opened 2025-07-03
fix #459: Set __STRICT_ANSI__=1 for non-gnu standards

### LINKED ISSUE #459: Set `__STRICT_ANSI__=1` for non-gnu standards

ludviggunne:


firewave:
> @firewave does this look OK to you?

Looks like it. I had not replied yet because I wanted to think about the awkwardness in the handling but I think that should go away when the defined are being re-done - see #386 and #387.

PR #463, opened 2025-07-03
fixed #462 - added CLI option `-l` to print line numbers

### LINKED ISSUE #462: add CLI option to print line numbers
Similar to Cppcheck it would be helpful if there was a way to display the line number in the output. This would help with the readability of code which is mostly macros.

firewave:


firewave:
Lacks the marker for line 1.

PR #479, opened 2025-08-04
fixed #478 - fail builds in CI on compiler warnings

### LINKED ISSUE #478: fail CI builds on compiler warnings
We should all builds in the CI when a compiler warning is encountered.

firewave:


firewave:
Requires #481 to be merged first.

PR #480, opened 2025-08-04
Makefile: added `CXXOPTS` and `LDOPTS` to extend `CXXFLAGS` and `LDFLAGS`

firewave:
Currently it was necessary to duplicate (and keep in sync) all the built-in flags if you just wanted to extend it. This introduced additional variables to make this easier.


firewave:
I will also look at adapting this for Cppcheck.

danmar (approved):

PR #488, opened 2025-08-11
Fix cppcheck #14050 Hang/memory exhaustion with circular includes

chrchr-github:
Not sure about the logic and how to test this.


firewave:
This should get a separate issue in the simplecpp repo - even if currently not reproducible.

chrchr-github:
Closed in favor of #497

Issue #424, opened 2025-04-07
do not depend on lifetime of user object in `TokenList`


firewave:
`TokenList` depends on a mutable `std::vector<std::string>` object which holds the files of files which have been tokenized. This seems dangerous as it ties to the lifetime of another object. In some cases this vector is not even necessary to be owned by the user since it is not used outside of the `TokenList`. It looks the list this could just be owned by `TokenList`

This supported by the fact that the copy constructors and assignment operators copy the reference which means that two different instances will modify that "global" vector which is obviously wrong and will cause issues.

There was some usage which provides the input file via that vector parameter instead of (or additionally) to the filename parameter. This allows to tokenize multiples explicitly but I am not sure if that is even necessary.

PR #493, opened 2025-08-14
Replace reference members with pointers

glankk:
Having references as class members complicates copy/move constructors and assignment operators and in some cases prevents them from being defaulted. They are also a source of bugs, as there are currently several instances of copy/move operators assigning to their referred object where the intended behavior most likely is to rebind them. This PR replaces most reference members with pointers, and makes explicit some of the semantics of affected classes with an explicitly deleted/defaulted copy/move assignment constructor/operator.

Fixes clang-tidy issues in #421.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -417,10 +417,16 @@ public:
         init();
     }
 
+    FileStream(const FileStream &other) = delete;
+    FileStream(FileStream &&other) = delete;
+
     ~FileStream() override {
         fclose(file);
         file = nullptr;
     }
+
+    FileStream &operator=(const FileStream &other) = delete;
+    FileStream &operator=(FileStream &&other) = delete;
 
     virtual int get() override {
         lastStatus = lastCh = fgetc(file);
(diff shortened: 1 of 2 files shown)


danmar:
I believe it is best practice to prefer reference instead of pointer;
https://isocpp.org/wiki/faq/references#refs-vs-ptrs

> Having references as class members complicates copy/move constructors and assignment operators and in some cases prevents them from being defaulted.

Imho, not if the intention is that it should still reference the same global object. We want that "files" point at the same object.

> They are also a source of bugs, as there are currently several instances of copy/move operators assigning to their referred object where the intended behavior most likely is to rebind them.

but just changing from reference to pointer does not solve that issue

glankk:
> I believe it is best practice to prefer reference instead of pointer; https://isocpp.org/wiki/faq/references#refs-vs-ptrs

I agree that references are preferable for public interfaces or trivial classes when possible, however semantically sensible copy and move assignments aren't really possible with non-trivial classes that contain references. References are still used in public-facing functions, only the internals are changed.

> Imho, not if the intention is that it should still reference the same global object. We want that "files" point at the same object.

If we really want to have exactly one global `files` then I think the lesser evil would be to have a singleton.
 
> but just changing from reference to pointer does not solve that issue

The bugs I'm referring to are on line 513, 529 and 1512 in simplecpp.cpp. It seems pretty clear to me that intent is to rebind the references, which is the behavior one would expect from an assignment operation, but that behavior is only possible with pointers. It's not needed for all classes though, `Location` is an example of a trivial class where a reference could be preferred.

firewave:
Maybe this should be approached from the other end by resolving #424 first.

PR #495, opened 2025-08-15
fixed some `Variable copied when it could be moved` Coverity warnings

firewave:


firewave:
callgrind in CI

`g++` - `75,902,495` -> `75,902,619`
`clang++` - `71,213,207` -> `71,211,565`

PR #497, opened 2025-08-15
Fix infinite loop with circular includes

glankk:
Fix infinite loop in `simplecpp::load` with circular includes by removing scans of cached files.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -3109,14 +3109,12 @@ bool simplecpp::FileDataCache::getFileId(const std::string &path, FileID &id)
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


firewave:
Does this consider code like

```
#define EXPR_1
#include "expr_stuff.h"
#undef EXPR_1
#define EXPR_2
#include "expr_stuff.h"
#undef EXPR_2
```

danmar:
@firewave this is related to a customer problem we have. I am not allowed to share any code unfortunately. but I can say that there are 1000's of headers that are included and it takes several minutes just to perform preprocessing of the file.

glankk:
> Is this in any way related to https://trac.cppcheck.net/ticket/14050?

I just tested and it does fix trac #14050 as well. It's an issue in simplecpp but it's only observed in cppcheck because the simplecpp CLI never invokes `load`.

firewave:
The PR title could be more descriptive.

danmar (approved):

PR #499, opened 2025-08-15
avoid direct usage of `std::cout`

firewave:


danmar (commented):
hmm these functions are for usage in the debugger. will I be able to call these during debugging? I guess std::cout symbol is not available while I debug inside the simplecpp.

We don't need to have these functions in release builds? would that solve your itch?

firewave:
> Integrating the proposed API while maintaining the possibility to just call the function seems like a reasonable trade-off to finalize this.

Like I said that would introduce the `<iostream>` header which it is one of the heavier headers that and as it is just debug code that seems unnecessary to spill into user code.

Also we could use default parameters which would require less code.

PR #501, opened 2025-08-17
fixed #500 - added callgrind step to CI

### LINKED ISSUE #500: add a callgrind step
Similar to Cppcheck we should add a callgrind step to the CI which processes a fixed input so we can track performance.

firewave:


.github/workflows/CI-unixish.yml
```diff
@@ -110,3 +110,22 @@ jobs:
       run: |
         make clean
         make -j$(nproc) test selfcheck CXXOPTS="-O2 -g3 -stdlib=libc++ -fsanitize=memory" LDOPTS="-lc++ -fsanitize=memory"
+
+    - name: Run callgrind
+      if: matrix.os == 'ubuntu-24.04'
+      run: |
+        wget https://github.com/danmar/simplecpp/archive/refs/tags/1.5.1.tar.gz
```
danmar, line 117: why don't we use latest HEAD? Does this create some baseline that you use for some comparison?

firewave:
Anything still to do here?

PR #505, opened 2025-08-23
improved testing of `#line` handling

firewave:


danmar:
> I also wanted to add tests for #file and #endfile but I cannot find any documentation on that and neither of the GNU compilers nor Visual Studio support it.

I invented it for internal use in cppcheck

danmar (approved):

PR #506, opened 2025-08-23
test.cpp: do not unconditionally remove comments

firewave:


firewave:
> Requires #507 to be fixed first.

Can be fixed later on. We can just remove them explicitly for now. No need to accumulate more technical debt.

danmar (approved):

PR #508, opened 2025-08-25
Improve runastyle scripts and update astyle version

jcfr:
This pull request improve the runastyle scripts for easier use:

- Ensure the scripts can be executed from any location.
- Centralize astyle configuration with a `.astylerc` file.
- Require astyle 3.4.13 (also available on PyPI) to preserve current formatting; newer 3.6.9 alters continuation-line indentation.
- Add support for command runners like `pipx` and `uvx` if `astyle` is not available in the `PATH`.
- Simplify the batch script with "@echo off" and proper error handling.
- Combine astyle invocations for `.h` and `.cpp` files into a single command.
- Format source files according to the updated style guidelines.

### DIFF
--- a/.astylerc
+++ b/.astylerc
@@ -1,0 +1,15 @@
+# style
+--style=kr
+--indent=spaces=4
+--indent-namespaces
+--lineend=linux
+--min-conditional-indent=0
+
+# options
+--pad-header
+--unpad-paren
+--suffix=none
+--convert-tabs
+--attach-inlines
+--attach-classes
+--attach-namespaces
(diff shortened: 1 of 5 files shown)


jcfr:
While working on finalizing https://github.com/danmar/simplecpp/pull/448, I realized that astyle integration could be further improved to simplify the developer experience.

Once this is integrated, I suggest we also introduce a workflow running astyle

jcfr:
Thanks for the review.

> so simplecpp should probably do the same.

Given limited bandwidth, I will not be able to switch to uncrustify. In the meantime, I think moving forward with this may be sensible as it allow contributor to apply the formatting.

danmar:
Thank you for this contribution but I fear that I feel we should switch to uncrustify instead asap. If you don't have the bandwidth I understand. But we more or less just need to copy the runformat script from the cppcheck repo and reuse all the options..

PR #509, opened 2025-08-26
Improve test runner script with refactoring and prerequisite checks

jcfr:
This pull request enhances the test runner script with two main improvements:

1. Refactors the script to introduce a `run()` function that centralizes the logic for executing compiler commands, capturing their output, and returning results.
2. Adds checks to ensure that required compilers (`clang`, `gcc`) are present before running tests.

### DIFF
--- a/run-tests.py
+++ b/run-tests.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import glob
 import os
 import shutil


run-tests.py
```diff
@@ -89,26 +119,12 @@ def cleanup(out):
...
-  comm = p.communicate()
-  clang_output = cleanup(comm[0])
+  clang_output = run(CLANG_EXE, cmd.split(' '))[1]
 
-  gcc_cmd = ['gcc']
-  gcc_cmd.extend(cmd.split(' '))
-  p = subprocess.Popen(gcc_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-  comm = p.communicate()
-  gcc_output = cleanup(comm[0])
+  gcc_output = run(GCC_EXE, cmd.split(' '))[1]
```
firewave, line 124:
```suggestion
  _, gcc_output, _ = run(GCC_EXE, cmd.split(' '))[1]
```

run-tests.py
```diff
@@ -78,6 +95,19 @@ def cleanup(out):
          'pr57580.c',
          ]
 
+
+def run(compiler_executable, compiler_args):
+  """Execute a compiler command and capture its output."""
+  compiler_cmd = [compiler_executable]
+  compiler_cmd.extend(compiler_args)
+  p = subprocess.Popen(compiler_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
firewave, line 103: While at it we should modernize this to `with ... as p:`

run-tests.py
```diff
@@ -78,6 +95,19 @@ def cleanup(out):
...
+  comm = p.communicate()
```
firewave, line 104: While at it we should use assign to `output, _`

run-tests.py
```diff
@@ -13,6 +14,22 @@ def cleanup(out):
...
+CLANG_EXE = shutil.which('clang')
+if not CLANG_EXE:
+    sys.exit('Failed to run tests: clang compiler not found')
+
+GCC_EXE = shutil.which('gcc')
+if not GCC_EXE:
+    sys.exit('Failed to run tests: gcc compiler not found')
+
+os.environ["PATH"] = os.getcwd() + os.pathsep + os.environ["PATH"]
+SIMPLECPP_EXE = shutil.which('simplecpp')
```
firewave, line 28: This is bad. You should not manipulate the path (that's what the `cwd` parameter of `subprocess.Popen()` - which is not needed in this case). And `which` is obviously not the way to lookup the executable we just built and exists in CWD.

run-tests.py
```diff
@@ -1,17 +1,42 @@
...
 import sys
+from typing import List, Tuple
 
-def cleanup(out):
-  ret = ''
-  for s in out.decode('utf-8').split('\n'):
-    if len(s) > 1 and s[0] == '#':
+
+def cleanup(out: str) -> str:
+  """Strip preprocessor output by removing whitespace."""
```
firewave, line 12: This is comment is just wrong. The function drops lines prefixed with `#`

run-tests.py
```diff
@@ -78,6 +103,26 @@ def cleanup(out):
...
+
+  try:
+    with subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, encoding="utf-8") as process:
+      stdout, stderr = process.communicate()
+      exit_code = process.returncode
+  except FileNotFoundError as e:
+    # Compiler not found
+    return (127, "", f"{e}")
+  except Exception as e:
+    return (1, "", f"{e}")
```
firewave, line 119: This simulates the exitcodes the shell might generate which might be misleading - we should specify our own so they can be differentiated.

run-tests.py
```diff
@@ -1,17 +1,33 @@
+from __future__ import annotations
 
 import glob
 import os
+import shutil
 import subprocess
 import sys
+from typing import List, Tuple
```
firewave, line 8: This looks unnecessary.

jcfr:
Thanks for the taking the time to review :pray: 

All the review comments have been addressed. Once the GitHub workflow have been enabled and the test pass, I believe this will be ready for integration :rocket: 

cc: @danmar @firewave

run-tests.py
```diff
@@ -1,17 +1,32 @@
...
+
+def cleanup(out: str) -> str:
+  parts = []
+  for line in out.splitlines():
+    if len(line) > 1 and line[0] == '#':
       continue
-    s = "".join(s.split())
-    ret = ret + s
-  return ret
+    parts.append("".join(line.split()))
```
firewave, line 15: This is not supposed to be stripped. The empty lines this might remove are actually important because this is preprocessor output and that would change the locations of the code.

hjmjohnson:
@jcfr Thank you for making this MUCH easier to understand.

run-tests.py
```diff
@@ -1,17 +1,30 @@
...
 import subprocess
 import sys
 
-def cleanup(out):
-  ret = ''
-  for s in out.decode('utf-8').split('\n'):
-    if len(s) > 1 and s[0] == '#':
+
+def cleanup(out: str) -> str:
+  parts = []
```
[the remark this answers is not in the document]
jcfr: Proposed updates are being consistent with _established_ convention so I don't think this should be an issue.

firewave (approved):
LGTM. The stuff I wanted to check will need to be redone anyways, so no need to further look into that.

run-tests.py
```diff
@@ -1,17 +1,32 @@
...
+  for line in out.splitlines():
```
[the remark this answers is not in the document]
firewave: Now that I looked into my stuff it turns out the changed behavior is actually fixing the tests on MinGW as those produce output with CRLF.

Issue #283, opened 2022-12-24
differentiate local includes and system includes


firewave:
Currently `DUI` only contains `includePaths` which are used for any include no matter if local or system.

There's also no support for `-isystem` in the `simplecpp` command-line tool.

firewave:
Oh - it looks like the existing code is actually correct.

https://gcc.gnu.org/onlinedocs/cpp/Invocation.html#Invocation
```
Directories specified with -iquote apply only to the quote form of the directive, #include "file". Directories specified with -I, -isystem, or -idirafter apply to lookup for both the #include "file" and #include <file> directives.
```

PR #511, opened 2025-08-26
feat: Ordered search paths for `-I`, `-isystem`, `-F`, `-iframework` + Darwin framework support (with legacy `-I` back-compat)

jcfr:
This PR modernizes simplecpp's header lookup to more closely match GCC/Clang, while preserving backward compatibility for existing users.

### What's new

* **Typed, ordered search paths**

  * Introduce `DUI::SearchPath` with `PathKind`:

    * `Include` (for `-I`)
    * `SystemInclude` (for `-isystem`)
    * `Framework` (for `-F`, Darwin)
    * `SystemFramework` (for `-iframework`, Darwin)
  * If `DUI::searchPaths` is non-empty, it is honored **verbatim** in the order provided.
    If it's empty, legacy `includePaths` are mirrored as `Include` entries for back-compat.

* **CLI flags**

  * Add support for `-F<dir>` and `-iframework<dir>` (framework lookups).
  * Add support for `-isystem<dir>` (system include lookups).
  * Keep `-I<dir>` as before.

* **Public API convenience**

  * New helpers on `DUI`:

    * `addIncludePath(path, legacy=false)`
    * `addSystemIncludePath(path)`
    * `addFrameworkPath(path)`
    * `addSystemFrameworkPath(path)`
  * `addIncludePath` defaults to `legacy=false`; tests exercise both modes.

* **Darwin frameworks**

  * Framework includes like `<Pkg/Hdr.h>` are resolved to `Pkg.framework/{Headers,PrivateHeaders}/Hdr.h`.
  * Implement `toAppleFrameworkRelatives()` returning prioritized candidates (Headers first, then PrivateHeaders).
  * Works for both `__has_include` and normal `#include`.

* **Lookup order (summary of actual implementation)**

  1. For `"quotes"` includes only: **directory of the including file** (unchanged).
  2. **Interleaved `searchPaths`** in left-to-right CLI order:

     * `Include` (`-I`)
     * `Framework` (`-F`) — using the Headers/PrivateHeaders rewrite when applicable
  3. **System include paths** (`SystemInclude`, i.e., `-isystem`)
  4. **System framework paths** (`SystemFramework`, i.e., `-iframework`)
  5. (Standard library dirs are outside simplecpp's scope; `-idirafter`/`-iquote` are not implemented in this series.)

* **Core loader refactors**

  * `openHeader()` and `FileDataCache::get()` now consult typed, ordered paths and produce candidates accordingly.
  * When resolving framework headers, both `Headers` and `PrivateHeaders` are tried in that order.

### Tests & tooling

* **Integration tests**

  * New helpers to build fixtures:

    * generic headers, sources, and minimal framework trees.
  * `test_framework_lookup`: verifies `Headers` vs `PrivateHeaders` and `-F` vs `-iframework`.
  * `test_searchpath_order`: a single parametrized test that validates precedence across **`-I`, `-isystem`, `-F`, `-iframework`**, including interleaving and duplicates, asserting on the exact `#line` path chosen.
* **Unit tests (`test.cpp`)**

  * Framework handling covered for both `#include` and `__has_include`.
  * Back-compat exercised by running key tests twice (new ordered API and legacy mode via `addIncludePath(..., /*legacy=*/true)`).

--- 

Supersedes the following pull request:
* https://github.com/danmar/simplecpp/pull/448

---

Related issues:
* Fixes https://github.com/danmar/simplecpp/issues/283

### DIFF
--- a/simplecpp.h
+++ b/simplecpp.h
@@ -430,7 +430,7 @@ namespace simplecpp {
          *  If 'legacy' is true, the path is added to the 'includePaths' vector;
          *  otherwise, it is added to 'searchPaths' with 'PathKind::Include'.
          */
-        void addIncludePath(const std::string& path, bool legacy) {
+        void addIncludePath(const std::string& path, bool legacy=false) {
             if (legacy) {
                 includePaths.push_back(path);
             } else {


jcfr:
> Looks like this is currently executed for all code even if there is no apple framework is available which seems excessive.

Indeed, this was a deliberate choice anticipating the use of the tool in the context of cross compilation. Adding an option to enable/disable the behavior may be best.

jcfr:
@glankk  Thanks for the suggestion :pray: 

All the review comments have been addressed. Once the GitHub workflow have been enabled and the test pass, I believe this will be ready for integration :rocket:

we also included a comprehensive docstring explaining how to work with the `DUI` struct.

```
/**
 * Command line preprocessor settings.
 *
 * Mirrors typical compiler options:
 *   -D <name>=<value>       Add macro definition
 *   -U <name>               Undefine macro
 *   -I <dir>                Add include search directory
 *   -F <dir>                Add framework search directory (Darwin)
 *   -iframework <dir>       Add system framework search directory (Darwin)
 *   --include <file>        Force inclusion of a header
 *   -std=<version>          Select language standard (C++17, C23, etc.)
 *
 * Path search behavior:
 *   - If searchPaths is non-empty, it is used directly, preserving the
 *     left-to-right order and distinguishing between Include, Framework,
 *     and SystemFramework kinds.
 *   - If searchPaths is empty, legacy includePaths is used instead, and
 *     each entry is treated as a normal Include path (for backward
 *     compatibility).
 */
```

To improve the developer experience, we could also further extend the `DUI` API:

```
// Mirrors GCC/Clang -I <dir>
void addIncludePath(const std::string& p) {
    searchPaths.push_back({p, PathKind::Include});
}

// Mirrors GCC/Clang -F <dir>
void addFrameworkPath(const std::string& p) {
    searchPaths.push_back({p, PathKind::Framework});
}

// Mirrors GCC/Clang -iframework <dir>
void addSystemFrameworkPath(const std::string& p) {
    searchPaths.push_back({p, PathKind::SystemFramework});
}
```

cc: @danmar

jcfr:
Waiting this is integrated, we will stage those changes into a fork and move forward with vendoring those into `PythonQt`.

Related:
* https://github.com/MeVisLab/pythonqt/pull/271

firewave:
#283 possibly needs to be addressed as prerequisite of this as the include directories are currently not differentiated between system and "local" ones.

jcfr:
Thanks again @hjmjohnson for working on the initial patch and establishing the momentum :pray: 

Hopefully our contributions will be integrated shortly :crossed_fingers:

jcfr:
> please look at adding such helper functions. It sounds good to me.

I will add those

> re: support for `-system` 

I am also adding support for this along with integration test update. This issue is related:
- https://github.com/danmar/simplecpp/issues/283

danmar:
it sounds like the simplecpp repo is not fully ready for this PR yet. but I hope we can get this PR into simplecpp repo soonish and then make a simplecpp release so we can upstream it into cppcheck and test it with daca etc.

there are several simplecpp fixes that will be nice to release with cppcheck 2.19.. as soon as it's tested properly etc..

firewave:
> I don't know how much this will actually affect the performance though, maybe it's negligible.

After #438 has been merged and I will extend the callgrind step with that and that *might* give an indication of the performance impact (so more to merge beforehand?).

> but I hope we can make the repo ready and get this PR into simplecpp repo soonish and then make a simplecpp release so we can upstream it into cppcheck and test it with daca etc.

As mentioned above we probably need to release a version to downstream before this is merged so we can prepare yet another patch (which also needs some test-related backports which have been requested by a packager).

simplecpp.cpp
```diff
@@ -2999,12 +3021,61 @@ static std::string openHeader(std::ifstream &f, const simplecpp::DUI &dui, const
...
-        if (!path.empty())
-            return path;
+    // Build an ordered, typed path list:
+    // - Prefer DUI::searchPaths when provided (interleaved -I/-F/-iframework).
+    // - Otherwise mirror legacy includePaths into Include entries (back-compat).
+    std::vector<simplecpp::DUI::SearchPath> searchPaths;
+    if (!dui.searchPaths.empty()) {
+        searchPaths = dui.searchPaths;
+    } else {
+        searchPaths.reserve(dui.includePaths.size());
```
danmar, line 3031: if `dui.searchPaths` is empty then we only need a short simple for loop here as far as I see.

main.cpp
```diff
@@ -43,17 +43,29 @@ int main(int argc, char **argv)
             }
             case 'I': { // include path
                 const char * const value = arg[2] ? (argv[i] + 2) : argv[++i];
-                dui.includePaths.push_back(value);
+                dui.addIncludePath(value, /* legacy= */ false);
```
danmar, line 46: if "legacy" is used simplecpp will not just skip over the new handling it will copy all the include paths (and as far as I see this is repeated). This solution will introduce a performance penalty.

We have had major performance problems and we managed to improve the situation but this I fear adds cpu overhead.

I envision that in Cppcheck many users will only use `-I` flags and I would like to be able to get optimal performance by choosing "legacy=true"

simplecpp.cpp
```diff
@@ -3051,32 +3123,78 @@ std::pair<simplecpp::FileData *, bool> simplecpp::FileDataCache::get(const std::
...
-            if (ret.first != nullptr) {
-                return ret;
+    // Build an ordered, typed path list:
+    // - Prefer DUI::searchPaths when provided (interleaved -I/-F/-iframework).
+    // - Otherwise mirror legacy includePaths into Include entries (back-compat).
+    std::vector<simplecpp::DUI::SearchPath> searchPaths;
+    if (!dui.searchPaths.empty()) {
+        searchPaths = dui.searchPaths;
+    } else {
+        searchPaths.reserve(dui.includePaths.size());
```
danmar, line 3141: this is a slowdown I fear. If searchPaths is empty we can speedup.

danmar (commented):
goal: try to make sure that performance is not affected negatively by this PR if we only provide `-I` flags as before..

danmar:
>> I don't see why a mingw release would be needed first?
> Since it exposed major bugs - it should have been ported back to 2.18 but there were just too many issues (it just exposed another one in the improved isAbsolutePath()).

People are waiting for the new simplecpp. It solves critical performance issues for a big customer.

I will tag simplecpp now so we can merge simplecpp to cppcheck.. and I feel that if mingw is not added soonish then let's merge this PR without it.

firewave:
> I will tag simplecpp now so we can merge simplecpp to cppcheck.. and I feel that if mingw is not added soonish then let's merge this PR without it.

The previous MinGW issues are all fixed but the absolute path tests exposed new issues - see #556. It is probably an easy fix but I need to do some manual testing to determine the current behavior and on how to test it properly in the CI. Since my cold has actually worsened I have not been able to look into it yet.

firewave:
> > well it would be nice with some callgrind scripts
> 
> There already is one but it does not rely on include paths as the additional selfcheck does.
> 
> > I believe we can use callgrind anyway to see the impact of this PR even if we don't have a script
> 
> Just add `valgrind --tool=callgrind` to the `./simplecpp` calls in `selfcheck.sh`. I have it parameterized and such and it just needs to be cleaned up so not much work is necessary I just cannot be bothered right now _cough_.

This is ready for review in #560.

jcfr:
Once #560 is integrated, I will rebase this pull request.

firewave:
> Once #560 is integrated, I will rebase this pull request.

It has been merged.

PR #514, opened 2025-08-28
clang-tidy.yml: updated to Clang 22

firewave:


simplecpp.cpp
```diff
@@ -3592,8 +3612,9 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
                 }
 
                 bool conditionIsTrue;
-                if (ifstates.top() == AlwaysFalse || (ifstates.top() == ElseIsTrue && rawtok->str() != ELIF))
+                if (ifstates.top() == AlwaysFalse || (ifstates.top() == ElseIsTrue && rawtok->str() != ELIF)) {
```
danmar: these `{}` do not feel super important to me. I think it's quite readable without.
this modification feels more related to uncrustify than clang-tidy, it's purely about coding style. is code updated automatically somehow?

PR #515, opened 2025-08-28
style: Format sources using astyle 3.0.1

jcfr:
This pull request reformats the codebase using **astyle 3.0.1** to ensure consistency with the project’s current style guidelines.

### Context

As noted in the review (https://github.com/danmar/simplecpp/pull/508#issuecomment-3234137957), the long-term plan is to transition formatting to **uncrustify** (likely by adapting the `runformat` script from the cppcheck repo). While that transition is being worked out, this change helps keep the repository consistent and avoids confusion for contributors.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -761,18 +761,18 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
             while (stream.good() && ch != '\n') {
                 currentToken += ch;
                 ch = stream.readChar();
-                if(ch == '\\') {
+                if (ch == '\\') {
                     TokenString tmp;
                     char tmp_ch = ch;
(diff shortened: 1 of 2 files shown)


jcfr:
Thanks for taking the time to follow up :pray: 

Looking forward to the improved code formatting infrastructure.

If possible consider:
- adding a `pre-commit.yml` GitHub workflow
- adding [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view) to ensure the stylistic changes do not pollute the history
- (ideally) having the corresponding binary available through tooling like `uvx` or `pipx`.


> it just creates extra stylistic changes in the git log.

I should have specified that this type of changes should be added to `.git-blame-ignore-revs` to address this.

PR #521, opened 2025-08-29
added support for `c2y` and `gnu2y`

firewave:


firewave:
This is to support `-std` options with such a value in projects. Since compilers have been supporting it for quite a while now it is not unlikely we might encounter it.

danmar (approved):

PR #523, opened 2025-08-29
do not use stream to read file in `FileDataCache::tryload()`

firewave:


firewave:
Using Callgrind with the extended selfcheck from #438:

`70,470,996` -> `69,235,694` - without system include
`1,474,800,488` -> `1,259,681,515` - with system include

CC @glankk

glankk (approved):

PR #525, opened 2025-08-30
simplecpp.h: prevent internal defines from spilling

firewave:


firewave:
> On a side note - the implementation-specific details which `SIMPLECPP_WINDOWS` is used for should not be exposed. That would also get rid of the need to include `windows.h` before `simplecpp.h` which is rather bad.

CC @glankk

PR #529, opened 2025-09-01
enabled and fixed `modernize-pass-by-value` clang-tidy warnings / avoid some unnecessary copies

firewave:


firewave:
callgrind in CI

`g++` - `74,990,727` -> `76,190,727`
`clang++` - `70,436,671` -> `71,500,835`

firewave:
Trying to mitigate the regression makes things worse although I am able to reduce the amount of `std::string` copies. There is still `memcpy()` calls I do not even understand where they are coming from.

PR #475, opened 2025-08-04
added MinGW workflow

firewave:

PR #531, opened 2025-09-02
Use `_WIN32` only to check for windows

glankk:
Removes `__CYGWIN__` and `__MINGW32__` checks, as well as the `SIMPLECPP_WINDOWS` macro, in favor of `_WIN32`. Needed for #475.


firewave:
And please mention the ticket it fixes in the title.

danmar:
Yes.. in general it is better to use `_WIN32` to check if windows is used.

PR #533, opened 2025-09-06
added expected output for selfcheck

firewave:


firewave:
This requires fixed input which isn't the case right now. Will be available after other selfcheck improvements.

PR #547, opened 2025-09-24
fixed #386 - rewrote define handling

### LINKED ISSUE #386: order of defines not respected
```cpp
void f()
{
#ifdef DEF_1
    int i;
#endif
}
```

```
$ ./simplecpp -DDEF_1 -UDEF_1 -DDEF_1 test.cpp
void f ( )
{



}
```

firewave:


firewave:
We need to keep (but deprecate) the old define handling for now.

PR #548, opened 2025-09-24
added `constness_ptr` as pointer wrapper to ensure actual method constness

firewave:


firewave:
I might spin out the class into a separate repo (as public domain) as this seems like something useful in general.

firewave:
This fails to compile with GCC before 15:
```
simplecpp.cpp:1742:37: error: operands to ‘?:’ have different types ‘const simplecpp::constness_ptr<simplecpp::Token>’ and ‘std::nullptr_t’
 1742 |                 valueToken = argtok ? argtok->next : nullptr;
      |                              ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

And with Visual Studio:
```
     6>C:\a\simplecpp\simplecpp\simplecpp.cpp(1742,54): error C2446: ':': no conversion from 'nullptr' to 'const simplecpp::constness_ptr<simplecpp::Token>' [C:\a\simplecpp\simplecpp\simplecpp_obj.vcxproj]
             C:\a\simplecpp\simplecpp\simplecpp.cpp(1742,54):
             Constructor for class 'simplecpp::constness_ptr<simplecpp::Token>' is declared 'explicit'
```

firewave:
We do not actually need to actually implement the class. As this is a compile-time check it would be enough to have a simple wrapper which does nothing and having a compile-time switch which makes it an alias for `std::experimental::propagate_const`.

PR #554, opened 2025-09-30
enabled `clang-analyzer-*` clang-tidy checks

firewave:
explicitly enable them because they will be disabled by default in a future version


firewave:
Obviously way too slow and noisy. Should be a scheduled job like in Cppcheck.

PR #582, opened 2025-11-01
Fix #581 (Failure when header is included twice with different macros defined)

### LINKED ISSUE #581: Failure when header is included twice with different macros defined
file1.c:
```
# define Y
# include "test.h"

# define BBB
# include "test.h"
```

test.h:
```
#if defined AAA
#elif defined BBB
# undef BBB
#endif

#ifdef BBB
# error adf
#endif
```

Output:
```
$ ~/simplecpp/simplecpp file.c
test.h:8: #error: #error adf
```

danmar:
There was a regression introduced by #468 . This PR just reverts the #468 changes and introduces a test.

### DIFF
--- a/integration_test.py
+++ b/integration_test.py
@@ -446,3 +446,38 @@ def test_incpath_dir(record_property, tmpdir):
 
     assert '' == stderr
     assert "error: could not open include 'inc'\n" == stdout
+
+
+def test_include_header_twice(tmpdir):
+    """ Issue #581 - Failure when header is included twice with different
+    macros defined"""
+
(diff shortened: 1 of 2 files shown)


danmar:
@firewave This is a critical bug for me and I can't release Cppcheck with this bug.

So when this fix is merged I want to make a new simplecpp tag and merge it to cppcheck.. you made many refactorings since 1.6.1 do you think we should tag HEAD or make a release branch that only contains this change..

olabetskyi (approved):

firewave:
CC @glankk

Issue #507, opened 2025-08-23
some preprocessor directives do not handle comments properly


firewave:
The test is unconditionally removing all comments (addressed in #506) leading to tests which test the behavior of such not actually testing it since they did not materialize in the token list.

Affected tests:
`has_include_2`
`missingHeader4`

It also affects the `#line lineno` preprocessor handling.

PR #586, opened 2025-11-05
fixed #504 / refs #507 - cleaned up and improved parsing of `#line` preprocessor directive

### LINKED ISSUE #504: Support prefix-less `#line` preprocessor directive without file
The GNU compilers also support a line preprocessor directive with just a number without a file specified.

```
# 3
__LINE__ __FILE__
```

`gcc -E output`:
```
# 0 "a.cpp"
# 0 "<built-in>"
# 0 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 0 "<command-line>" 2
# 1 "a.cpp"
# 3 "a.cpp"
3 "a.cpp"
```

actual:
```

2 "a.cpp"
```

firewave:


simplecpp.cpp
```diff
@@ -696,42 +696,44 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
...
-                    }
-                }
+
+                if (ppTok->str() == "line")
+                    ppTok = advanceAndSkipComments(ppTok);
+
+                if (!ppTok || !ppTok->number)
+                    continue;
+
+                const unsigned int line = std::atol(ppTok->str().c_str());
```
danmar, line 726: could you please add a (portability) warning and some handling for wrong values? I believe negative values for instance are undefined behavior. The C99 specification says that it must be "digit-sequence". And:

> the digit sequence shall not specify zero, nor a number greater than 2147483647.

firewave: I guess you mean a warn if the GNU extension version is being used. I think if the non-supporting compilers (i.e. MSVC) reject it that is unnecessary.

Regarding the handling of more (any) invalid values I would like to handle that through a separate issue and PR as a follow-up.

danmar: > Regarding the handling of more (any) invalid values I would like to handle that through a separate issue and PR as a follow-up.

ok

danmar (approved):

simplecpp.cpp
firewave: I filed #663 about it.

PR #599, opened 2025-12-08
improved `simplecpp::TokenList` constructors

firewave:
- added `simplecpp::View` to be able to write code which is compatible with C++11 and => C++17
- make constructors which take an array always available - since the size has not to be provided by the user it can be considered "safe"
- added tests for all inputs

### DIFF
--- a/simplecpp.h
+++ b/simplecpp.h
@@ -257,7 +257,6 @@ namespace simplecpp {
         explicit TokenList(std::vector<std::string> &filenames);
         /** generates a token list from the given std::istream parameter */
         TokenList(std::istream &istr, std::vector<std::string> &filenames, const std::string &filename=std::string(), OutputList *outputList = nullptr);
-#ifdef SIMPLECPP_TOKENLIST_ALLOW_PTR
         /** generates a token list from the given buffer */
         template<size_t size>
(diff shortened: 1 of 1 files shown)


test.cpp
```diff
@@ -3299,11 +3299,44 @@ static void preprocess_files()
     }
 }
 
-static void safe_api()
+static void tokenlist_api()
 {
+    std::vector<std::string> filenames;
+    {
+        char input[] = "code"; // NOLINT(misc-const-correctness)
```
firewave: This is not actually reported but per-emptively added in case it is detected in the future. I added a comment about it upstream: https://github.com/llvm/llvm-project/issues/56758#issuecomment-3638935265

simplecpp.h
```diff
@@ -69,6 +69,46 @@ namespace simplecpp {
     enum cppstd_t : std::int8_t { CPPUnknown=-1, CPP03, CPP11, CPP14, CPP17, CPP20, CPP23, CPP26 };
 
     using TokenString = std::string;
+
+#if defined(__cpp_lib_string_view) && !defined(__cpp_lib_span)
```
danmar: what are these macros? are these standard macros? spontanously I think it might be an alternative to check some standard macro that gives us the c++ version ?

PR #628, opened 2026-02-24
Fix #613: support binary literals in preprocessor checks

### LINKED ISSUE #613: support binary literals in preprocessor checks
```cpp
#if A == 0b1
int i;
#endif
```

There is no code using `-DA1`.

Binary literals were introduced in C++14. This is supported by all major compilers.

ludviggunne:


firewave:
Please rebase.

simplecpp.cpp
```diff
@@ -152,7 +152,8 @@ static unsigned long long stringToULL(const std::string &s, std::size_t offset =
         } else if ('A' <= c && c <= 'F') {
             d = 10 + c - 'A';
         } else {
-            throw std::runtime_error("invalid integer literal");
+            // TODO: handle suffix
```
firewave: Please add a ticket if do not plan to address this in this PR.

ludviggunne: I think a pretty big change is needed to handle integer conversions/overflow correctly (such as in `expr_usual_conversions.c`). So I could either keep that file excluded for this PR or tackle the bigger issue before this. Either way I'll create an issue.

ludviggunne: https://github.com/cppcheck-opensource/simplecpp/issues/675

I couldn't find a way to add a label, maybe I don't have permissions

PR #636, opened 2026-03-27
Fix hang on UTF-16 LE BOM file

chrchr-github:
fixes #637

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -285,7 +285,7 @@ public:
         // character is non-ASCII character then replace it with 0xff
         if (isUtf16) {
             (void)get();
-            const int ch2 = static_cast<unsigned char>(peek());
+            const auto ch2 = static_cast<unsigned char>(peek());
             unget();
             const int ch16 = makeUtf16Char(ch, ch2);
             ch = static_cast<unsigned char>(((ch16 >= 0x80) ? 0xff : ch16));
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -276,16 +276,18 @@ public:
     }
 
     unsigned char peekChar() {
-        auto ch = static_cast<unsigned char>(peek());
+        const int pk = peek();
+        auto ch = static_cast<unsigned char>(pk);
+        if (pk == EOF)
+            return ch;
 
         // For UTF-16 encoded files the BOM is 0xfeff/0xfffe. If the
         // character is non-ASCII character then replace it with 0xff
         if (isUtf16) {
             (void)get();
-            const int ch2 = peek();
-            if (ch2 != EOF)
-                unget();
-            const int ch16 = makeUtf16Char(ch, static_cast<unsigned char>(ch2));
+            const int ch2 = static_cast<unsigned char>(peek());
+            unget();
+            const int ch16 = makeUtf16Char(ch, ch2);
             ch = static_cast<unsigned char>(((ch16 >= 0x80) ? 0xff : ch16));
         }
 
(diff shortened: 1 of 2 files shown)


firewave:
With #647 it should be possible to write a test which uses the input in question.

chrchr-github:
> With #647 it should be possible to write a test which uses the input in question.

Can we just run the tests with all three options?

firewave:
> Can we just run the tests with all three options?

Yes, if somebody adds the functionality. It didn't seem straight forward to me so I haven't yet. Feel free to do so.

firewave:
I think I took a wrong turn. That should trigger with regular unit test. So please add a test for it.

firewave:
> I think I took a wrong turn. That should trigger with regular unit test. So please add a test for it.

Ah - you mentioned `FileStream` in the ticket. Still I think we should have a unit test (if missing).

simplecpp.cpp
```diff
@@ -276,7 +276,10 @@ class simplecpp::TokenList::Stream {
     }
 
     unsigned char peekChar() {
-        auto ch = static_cast<unsigned char>(peek());
+        const int pk = peek();
+        auto ch = static_cast<unsigned char>(pk);
+        if (pk == EOF)
+            return ch;
```
danmar, line 281: I appreciate that we handle EOF. However, spontanously it does not seem optimal to return some abitrary ch value. It's implementation defined what value of EOF is right? If we would return 0xFF we at least know what the return value is. Or do we want to have the implementation defined return value?

firewave: This needs to return `EOF` since we are a custom layer on top of `getc()`. And since we use the `EOF` macro it is not "arbitrary" but the one defined by the libc implementation.

danmar: well.. we are not returning EOF. That is why I said it returns a arbitrary value. I.e. if EOF is FFFFFF41 then we return a `'A'` char (0x41) here which is not equal to EOF.

danmar: > This needs to return EOF since we are a custom layer on top of getc().

spontanously this sounds best to me.

chrchr-github: `EOF` is `int`, but the return type is `unsigned char`. We already return `EOF` cast to `unsigned char`.

danmar: It would be more proper that `peekChar()` returns the int value.

If we assume that EOF is always -1 then as far as I see our usage will be safe currently. However it could become unsafe in the future.
```
    if (num && ch=='\'' && isNameChar(stream.peekChar()))
```
isNameChar returns false for FF and it should return false for EOF.

If we assume that EOF returns `'A'` on some machine then that code will suddenly not work properly. It will think that `'A'` is a name char.

chrchr-github: Done.

simplecpp.cpp
```diff
@@ -598,7 +600,7 @@ std::string simplecpp::TokenList::stringify(bool linenrs) const
     return ret.str();
 }
 
-static bool isNameChar(unsigned char ch)
+static bool isNameChar(int ch)
 {
     return std::isalnum(ch) || ch == '_' || ch == '$';
```
danmar, line 605: I believe the input to std::isalnum must be 0-0xff

danmar: hmm the documentation says:

>  the behavior of std::isalnum is undefined if the argument's value is neither representable as unsigned char nor equal to EOF.

so we are safe right?

chrchr-github: > hmm the documentation says:
> 
> > the behavior of std::isalnum is undefined if the argument's value is neither representable as unsigned char nor equal to EOF.
> 
> so we are safe right?

That's how I read it too.

danmar (approved):

PR #639, opened 2026-04-03
Refs #638: Avoid shadowing arguments

ludviggunne:
Fixes `shadowFunction` that appears with https://github.com/danmar/cppcheck/pull/8303

### DIFF
--- a/simplecpp.h
+++ b/simplecpp.h
@@ -174,9 +174,9 @@ namespace simplecpp {
         bool isOneOf(const char ops[]) const;
         bool startsWithOneOf(const char c[]) const;
         bool endsWithOneOf(const char c[]) const;
-        static bool isNumberLike(const std::string& str) {
-            return std::isdigit(static_cast<unsigned char>(str[0])) ||
-                   (str.size() > 1U && (str[0] == '-' || str[0] == '+') && std::isdigit(static_cast<unsigned char>(str[1])));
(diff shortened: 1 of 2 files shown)


firewave:
> Let's wait a few more days to see if it re-appears, otherwise we disable it in the CI.

I disabled it - please rebase.

simplecpp.h
```diff
@@ -396,7 +396,7 @@ namespace simplecpp {
         void constFoldQuestionOp(Token *&tok1);
 
         std::string readUntil(Stream &stream, const Location &location, char start, char end, OutputList *outputList);
-        void lineDirective(unsigned int fileIndex, unsigned int line, Location &location);
+        void lineDirective(unsigned int fileId, unsigned int line, Location &location);
```
firewave, line 400: That change is a bit problematic because "fileIndex" has a special meaning within simplecpp and that would get lost...

firewave: A rather awkward solution would be using `fileIndex_`. Maybe somebody else could chime in...

firewave: And on side note - as index and ID are different things it should have been changed to `fileIdx`.

danmar: I agree it would be good to keep the "index" notion in some way.

ludviggunne: I'm fine with the trailing underscore.

danmar (approved):

PR #645, opened 2026-04-10
main.cpp: fixed handling of options without value

firewave:


firewave:
This would have allowed options simply starting with the name to be allowed. That is fine for the `-D`, `-U`, `-I` options but not the others.

danmar (approved):

PR #647, opened 2026-04-12
main.cpp: added option `-input=file|fstream|fstream|buffer` to specify the `TokenList` interface / removed `-is`

firewave:


firewave:
Prerequisite for #625.

main.cpp
```diff
@@ -157,7 +179,10 @@ int main(int argc, char **argv)
         std::cout << "  -UNAME          Undefine NAME." << std::endl;
         std::cout << "  -std=STD        Specify standard." << std::endl;
         std::cout << "  -q              Quiet mode (no output)." << std::endl;
-        std::cout << "  -is             Use std::istream interface." << std::endl;
+        std::cout << "  -file           Consume input as file pointer (default)." << std::endl;
```
danmar: I don't have strong opinions but an alternative would be: ` -input=file|fs|ss|buf`

danmar (approved):
after considering my comment, feel free to merge. you can choose if we will have 1 argument or 4. I assume it will be possible to change it later if we want.

main.cpp
firewave: I like it. I based it on the existing single option and did not consider using a key-value approach.

main.cpp
```diff
@@ -86,9 +93,28 @@ int main(int argc, char **argv)
...
                     }
                     dui.includes.emplace_back(std::move(value));
-                } else if (std::strcmp(arg, "-is")==0) {
+                }
+                else if (std::strncmp(arg, "-input=",7)==0) {
                     found = true;
-                    use_istream = true;
+                    const std::string input = arg + 7;
+                    if (input.empty()) {
+                        std::cout << "error: option -inout with no value." << std::endl;
```
danmar:
```suggestion
                        std::cout << "error: option -input with no value." << std::endl;
```

danmar (approved):

danmar (approved):

PR #651, opened 2026-04-27
moved part of `FileDataCache` implementation into source file

firewave:


simplecpp.cpp
```diff
@@ -3147,7 +3209,14 @@ std::pair<simplecpp::FileData *, bool> simplecpp::FileDataCache::get(const std::
...
 
-bool simplecpp::FileDataCache::getFileId(const std::string &path, FileID &id)
+void simplecpp::FileDataCache::clear()
+{
+    mImpl->clear();
+    mNameMap.clear();
+    mData.clear();
+}
+
+bool getFileId(const std::string &path, FileID &id)
```
danmar: should it be `static` also? the declaration is static as far as I can tell.

firewave: Yes. That should have been caught by the compiler or clang-tidy though.

simplecpp.cpp
```diff
@@ -3075,6 +3078,65 @@ static std::string openHeader(std::ifstream &f, const simplecpp::DUI &dui, const
...
+simplecpp::FileDataCache::FileDataCache()
+    : mImpl(new Impl)
+{}
+
+simplecpp::FileDataCache::~FileDataCache() = default;
+simplecpp::FileDataCache::FileDataCache(FileDataCache &&) noexcept = default;
+simplecpp::FileDataCache &simplecpp::FileDataCache::operator=(simplecpp::FileDataCache &&) noexcept = default;
+
+static bool getFileId(const std::string &path, FileID &id);
+
```
firewave, line 3154: Ah - I forward declared it so the code doesn't need to be moved.

simplecpp.cpp
danmar: I am confused by your comments.. do you plan to add the static keyword or is there some reason to not add it?

firewave: It is already `static` in the forward declaration - see the other comment thread.

danmar: so the reason you don't write `static` here is because that would be redundant?

To me it would feel preferable to match the prototype and write the `static` keyword even if it's technically redundant.

firewave: Now I am finally getting what your are aiming for. I have never seen code written like that. Usually `static` (like `extern`) is only applied to the declaration. It is valid code though and does not even trigger a warning (which is surprising to me).

danmar: I think the reason the `extern` is usually not written is that it's totally redundant. the function is extern with or without the keyword.. 

I want that we have matching declaration and definition. I don't want that any `static` or typename or variablename differ or anything even if it does not technically matter.

danmar:
```suggestion
static bool getFileId(const std::string &path, FileID &id)
```
for methods the `static` must be left out for the out of line definition. but that doesn't mean it should be left out here.

firewave: Done.

danmar (approved):

PR #658, opened 2026-05-31
Fix #657: combineOperators(): &= gets split up incorrectly

ludviggunne:


wjakobsson (approved):

danmar (approved):

simplecpp.cpp
```diff
@@ -1040,7 +1025,7 @@ void simplecpp::TokenList::combineOperators()
                 continue;
             }
             const Token *prev = tok->previous;
-            while (prev && prev->isOneOf(";{}()"))
+            while (prev && prev->isOneOf(";{}("))
                 prev = prev->previous;
             executableScope.push(prev && prev->op == ')');
```
chrchr-github, line 1030: It seems like this is now equivalent to `push(false)`.

firewave: Not if the *initial* `tok->previous->op` is `)`.

PR #659, opened 2026-06-10
optimized `std::stack` usage in `simplecpp::TokenList::combineOperators()`

firewave:
- use `std::vector` as underlying container for `std::stack`
- initialize `std::stack` with initial value instead of inserting it

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1031,7 +1031,7 @@ static bool isAlternativeAndBitandBitor(const simplecpp::Token* tok)
 
 void simplecpp::TokenList::combineOperators()
 {
-    std::stack<bool> executableScope;
+    std::stack<bool, std::vector<bool>> executableScope;
     executableScope.push(false);
     for (Token *tok = front(); tok; tok = tok->next) {
         if (tok->op == '{') {
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1031,8 +1031,7 @@ static bool isAlternativeAndBitandBitor(const simplecpp::Token* tok)
 
 void simplecpp::TokenList::combineOperators()
 {
-    std::stack<bool, std::vector<bool>> executableScope;
-    executableScope.push(false);
+    std::stack<bool, std::vector<bool>> executableScope{{false}};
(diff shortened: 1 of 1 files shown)


simplecpp.cpp
```diff
@@ -1031,8 +1031,7 @@ static bool isAlternativeAndBitandBitor(const simplecpp::Token* tok)
 
 void simplecpp::TokenList::combineOperators()
 {
-    std::stack<bool> executableScope;
-    executableScope.push(false);
+    std::stack<bool, std::vector<bool>> executableScope{{false}};
```
danmar, line 1034: isn't `std::vector<bool>` bad to use overall in c++11..
are you sure that all usage here will work fine..

firewave: As this is about performance and valgrind indicates that is fine in either compiler so I see no problem.

libc++ has been adding special handling for `vector<bool>` and we are also not using it as a bit field here (which is the common usage for that IIRC).

danmar: I am not talking about performance nor stuff that valgrind will detect. I referred to such problems:
* No direct references: Because individual bits don't have memory addresses, std::vector<bool>::operator[] cannot return a bool&. Instead, it returns a temporary proxy object
* Breaks generic code: If you write a template function expecting a T&, it will fail to compile if T is bool because a proxy object cannot be bound as a standard reference
* Pointer failures: You cannot create a bool* pointing to elements inside a vector<bool> since a pointer requires an addressable byte.
* Concealed auto-deduction issues: If you use auto& to deduce a boolean element, you will actually deduce the proxy object, not a boolean.
* Thread safety pitfalls: Because multiple boolean elements share the same underlying byte or machine word, writing to different adjacent bits from different threads causes data races. They logically represent different indices, but they are physically the same memory block.
* Incompatibility with Modern APIs: Modern C++ features like std::span<bool> cannot be used directly with vector<bool> because there is no contiguous bool* storage to point at.

firewave: None of that applies here since we are using it through `std::stack` and only use it through that interface.

danmar (approved):

PR #661, opened 2026-06-10
avoid unnecessary container lookups in `simplecpp::preprocess()`

firewave:


danmar (approved):
this feels safer also. a typo would not have been detected by the compiler.

PR #670, opened 2026-07-02
Fix #669: Manually included files are not reported missing by preprocess

glankk:


danmar (commented):
hmm would it be possible to test something here.

danmar (approved):

PR #672, opened 2026-07-08
Fix #671: Missing newline at end of file should produce a warning

glankk:


main.cpp
```diff
@@ -275,6 +275,9 @@ int main(int argc, char **argv)
             case simplecpp::Output::PORTABILITY_BACKSLASH:
                 std::cerr << "portability: ";
                 break;
+            case simplecpp::Output::PORTABILITY_NO_EOF_NEWLINE:
+                std::cerr << "portability: ";
```
danmar, line 284: we could adjust main.cpp so that this output is only written if the filename extension is ".c".

glankk: I think this is a bit of a clumsy way to solve this issue, I know there exists production code where c++ files have a .c extension (gdb is one example of this). We could have command-line flag for choosing the standard to use for simplecpp, but I think it might be better to just suppress this in the simplecpp executable and let cppcheck choose how to handle this when thrown from the simplecpp library. As far as I know neither the gcc or clang preprocessor have a warning for this, so there's precedence.

glankk: Sorry, I just realized we do have a -std flag for simplecpp. We should probably use that to decide what to do with this warning, and use the filename as a backup for autodetecting the standard to use.

danmar: main.cpp is an "example implementation" more or less. It doesn't have to be perfect. If the `-std` option has been specified then please rely on that primarily.

Cppcheck has both `--std` and `--language`. It has better knowledge about the language than simplecpp.
The `--language` option is not communicated to simplecpp.

glankk: I suggest something like this:
```
case simplecpp::Output::PORTABILITY_NO_EOF_NEWLINE:
  if (simplecpp::getCStd(dui.std) == simplecpp::CUnknown) {
      // Only UB for c code, suppress for c++ code
      // If no standard is specified then prefer to have a false negative
      continue;
  }
  std::cerr << "portability: ";
  break;
```
cppcheck will have it's own way of handling this obviously, but I think this is okay for the simplecpp executable.

danmar: > I know there exists production code where c++ files have a .c extension

yes sure. And in these cases you would use the cppcheck option `--language=c++` .

If you just execute `cppcheck --std=c++11 path` then you say that c++ files in the path should be analyzed using "c++11". and c files in the path are analyzed with default C standard.

danmar (approved):

PR #681, opened 2026-07-13
Fix #680: Add option to keep comments in preprocessed code

ludviggunne:


danmar (commented):
I don't see the use case to treat comments differently in the source file and in included files. do you know the reason?

test.cpp
```diff
@@ -562,6 +565,27 @@ static void comment_multiline()
...
+    }
+
+    {
+        const char code[] = "// comment\n";
+        ASSERT_EQUALS("// comment", preprocess(code,dui));
+    }
+
+    {
+        const char code[] = "#define MACRO /* comment */\nMACRO\n";
+        ASSERT_EQUALS("\n/* comment */", preprocess(code,dui));
```
danmar, line 585: this may be questionable. do we want that comments are expanded. maybe the comment should only be kept on line 1.

but for now .. as far as I see it does not matter much. If somebody has strong opinions later and wants to change I am not against changing..

danmar (approved):

test.cpp
ludviggunne: gcc does copy comments in expansions. it will however turn c++-style comments in to c-style comments: https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html

PR #687, opened 2026-07-22
Fix #686: Add warning for bad line directives

glankk:
`readfile` now takes a `DUI` with information about which standard to use, which is needed to determine if a warning should be issued or not. `preprocess` now also includes outputs from `makeTokenLists` which calls `readfile`. Some tests had to be fixed to have an ending newline because of this in order to pass.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -4026,7 +4026,7 @@ void simplecpp::cleanup(FileDataCache &cache)
     cache.clear();
 }
 
-simplecpp::cstd_t simplecpp::getCStd(const std::string &std, cstd_t dflt)
+simplecpp::cstd_t simplecpp::getCStd(const std::string &std)
 {
     if (std == "c90" || std == "c89" || std == "iso9899:1990" || std == "iso9899:199409" || std == "gnu90" || std == "gnu89")
         return C89;
@@ -4040,7 +4040,7 @@ simplecpp::cstd_t simplecpp::getCStd(const std::string &std, cstd_t dflt)
         return C23;
     if (std == "c2y" || std == "gnu2y")
         return C2Y;
-    return dflt;
+    return CUnknown;
 }
 
 const char *simplecpp::getCStdName(cstd_t std)
@@ -4090,7 +4090,7 @@ std::string simplecpp::getCStdString(const std::string &std)
     return getCStdString(getCStd(std));
 }
 
-simplecpp::cppstd_t simplecpp::getCppStd(const std::string &std, cppstd_t dflt)
+simplecpp::cppstd_t simplecpp::getCppStd(const std::string &std)
 {
     if (std == "c++98" || std == "c++03" || std == "gnu++98" || std == "gnu++03")
         return CPP03;
@@ -4106,7 +4106,7 @@ simplecpp::cppstd_t simplecpp::getCppStd(const std::string &std, cppstd_t dflt)
         return CPP23;
     if (std == "c++26" || std == "c++2c" || std == "gnu++26" || std == "gnu++2c")
         return CPP26;
-    return dflt;
+    return CPPUnknown;
 }
 
 const char *simplecpp::getCppStdName(cppstd_t std)
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -767,15 +767,16 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
                             msg += "Line number zero is undefined behavior.";
                         } else {
                             msg += "Line numbers above " + std::to_string(maxline) + " are ";
-                            if (unknown_std || cppstd >= CPP26)
+                            if (unknown_std)
+                                msg += "undefined behavior or conditionally supported";
+                            else if (cppstd >= CPP26)
                                 msg += "conditionally supported";
                             else
                                 msg += "undefined behavior";
-                            msg += " in ";
                             if (cstd != CUnknown)
-                                msg += getCStdName(cstd);
-                            else
-                                msg += getCppStdName(unknown_std ? CPP26 : cppstd);
+                                msg += " in " + getCStdName(cstd);
+                            else if (cppstd != CPPUnknown)
+                                msg += " in " + getCppStdName(cppstd);
                             msg += ".";
                         }
                         simplecpp::Output err{
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -755,8 +755,7 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
                         } else {
                             msg += "Line numbers above " + std::to_string(maxline) + " are " +
                                     (cppstd == CPP26 ? "conditionally supported" : "undefined behavior") +
-                                    " in " +
-                                    (std_is_c ? getCStdName(cstd) : getCppStdName(cppstd)) + ".";
+                                    " in " + (std_is_c ? getCStdName(cstd) : getCppStdName(cppstd)) + ".";
                         }
                         simplecpp::Output err{
                             simplecpp::Output::PORTABILITY_LINE_DIRECTIVE,
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -754,9 +754,9 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
                             msg += "Line number zero is undefined behavior.";
                         } else {
                             msg += "Line numbers above " + std::to_string(maxline) + " are " +
-                                (cppstd == CPP26 ? "conditionally supported" : "undefined behavior") +
-                                " in " +
-                                (std_is_c ? getCStdName(cstd) : getCppStdName(cppstd)) + ".";
+                                    (cppstd == CPP26 ? "conditionally supported" : "undefined behavior") +
+                                    " in " +
+                                    (std_is_c ? getCStdName(cstd) : getCppStdName(cppstd)) + ".";
                         }
                         simplecpp::Output err{
                             simplecpp::Output::PORTABILITY_LINE_DIRECTIVE,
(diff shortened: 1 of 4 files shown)


test.cpp
```diff
@@ -2847,6 +2857,115 @@ static void nullDirective3()
...
+                          toString(outputList));
+        }
+
+        {
+            const char code[] =
+                "#line 18446744073709551617\n" // 2^64 + 1
+                ";\n";
+            simplecpp::OutputList outputList;
+            makeTokenList(code, dui, &outputList);
+            ASSERT_EQUALS("file0,2,portability_line_directive,Line number out of range: 18446744073709551617.\n",
```
danmar: I feel that ideally the warning would write the range permitted by the standard. But I am not sure exactly how I want to say it..

Can the warning say that it's UB according to the standard , and write the used standard...

If latest C++ standard is used it would be nice to say that it is conditionally supported instead of UB. However I don't know if the C++ standard allows negative value at all maybe we should say that negative values are unspecified or something..

glankk: Negative values are not even allowed syntactically according to the standard, they should fail to compile due to syntax error.

glankk: My point being that the standard doesn't cover the case of negative numbers, so I think we can ignore them.

test.cpp
```diff
@@ -2847,6 +2857,191 @@ static void nullDirective3()
...
+            std_name = simplecpp::getCStdName(simplecpp::getCStd(std));
+        else
+            std_name = simplecpp::getCppStdName(simplecpp::getCppStd(std));
+
+        simplecpp::DUI dui;
+        dui.std = std;
+
+        {
+            const char code[] =
+                "#line 0\n"
```
danmar, line 2944: could we add a negative value test and check what the output is. syntax error or such portability warnings would be fine ... it's not possible to detect such code with cppcheck so I feel we should add this in simplecpp.

glankk: We could output a syntax error and stop processing when the value is negative, that would be the behavior ordained by the standard. What do you think?

glankk: On the other hand, simplecpp is generally pretty lenient when it comes to syntax errors, so it feels weirdly specific to add an error for this. For example, `#line 1 garbage here` is accepted without error.

glankk: In the current implementation, line directives with negative numbers are silently ignored, which I think is okay. Since the code shouldn't even compile according to the standard, you shouldn't really need cppcheck to find such errors. cppcheck doesn't generally check for syntax errors as far as I'm aware, since they're caught by the compiler.

danmar: I think it's good to be "somewhat lenient" when we can handle the syntax error in a OK way. Sometimes users wants that.. i.e. when compilers allows some code it's annoying if cppcheck stops analysis on that code..

But I fear that negative numbers could lead to more problems later in the client if we would allow it..

simplecpp.cpp
```diff
@@ -727,10 +737,47 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
...
+                }
+
+                if (line == 0 || line > maxline) {
+                    if (outputList) {
+                        std::string msg = "Line number out of range: " + ppTok->str() + ". ";
+                        if (line == 0) {
+                            msg += "Line number zero is undefined behavior.";
+                        } else {
+                            msg += "Line numbers above " + std::to_string(maxline) + " are " +
+                                   (cppstd == CPP26 ? "conditionally supported" : "undefined behavior") +
```
danmar: it's probably not undefined behavior again in future versions..
```suggestion
                                   (cppstd >= CPP26 ? "conditionally supported" : "undefined behavior") +
```

simplecpp.cpp
```diff
@@ -659,13 +659,23 @@ void simplecpp::TokenList::lineDirective(unsigned int fileIndex_, unsigned int l
...
+void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename, const DUI &dui, OutputList *outputList)
 {
     unsigned int multiline = 0U;
     bool trailing_nl = true;
 
     const Token *oldLastToken = nullptr;
 
+    const cstd_t cstd = getCStd(dui.std);
+    const bool std_is_c = cstd != CUnknown;
+    const cppstd_t cppstd = getCppStd(dui.std, std_is_c ? CPPUnknown : CPP26); // use C++26 by default
```
danmar: I feel uncomfortable about this code. You placed it here so the variables should have proper values for the whole scope not just when processing line directives..

It feels wrong in general in the simplecpp to use C++26 by default. If DUI::std is empty it might as well be C code.

I am not sure what I think should happen if DUI::std is empty.

simplecpp.cpp
```diff
@@ -659,13 +659,23 @@ void simplecpp::TokenList::lineDirective(unsigned int fileIndex_, unsigned int l
...
+
+    unsigned long maxline;
```
danmar: We could move this `maxline` variable down to where it's used. It will not be a CPU bottle neck imo to calculate it for each line directive.

simplecpp.cpp
danmar: the natural code would be:
```suggestion
    const cstd_t cstd = getCStd(dui.std);
    const cppstd_t cppstd = getCppStd(dui.std);
```

danmar: and if you want to handle empty dui.std in some special way when handling line directives it makes more sense to put that handling where the line directives are processed.

glankk: I think it makes sense to assume the latest standard if no one is specified. It doesn't make sense to not assume a standard, because in that case it doesn't even make sense to have a warning in the first place because we don't know what the rules are. And if we're to assume a standard then it should probably either be the latest or the most lenient, in this case and probably most cases that means C++26.

simplecpp.cpp
```diff
@@ -727,10 +730,62 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
...
+                            if (unknown_std || cppstd >= CPP26)
+                                msg += "conditionally supported";
+                            else
+                                msg += "undefined behavior";
+                            msg += " in ";
+                            if (cstd != CUnknown)
+                                msg += getCStdName(cstd);
+                            else
+                                msg += getCppStdName(unknown_std ? CPP26 : cppstd);
+                            msg += ".";
```
danmar: how about:

```suggestion
                            if (unknown_std)
                                msg += "undefined behavior or conditionally supported";
                            else if (cppstd >= CPP26)
                                msg += "conditionally supported";
                            else
                                msg += "undefined behavior";
                            if (cstd != CUnknown)
                                msg += " in " + getCStdName(cstd);
                            else if (cppstd != CPPUnknown)
                                msg += " in " + getCppStdName(cppstd);
                            msg += ".";
```

simplecpp.cpp
```diff
@@ -3970,7 +4026,7 @@ void simplecpp::cleanup(FileDataCache &cache)
     cache.clear();
 }
 
-simplecpp::cstd_t simplecpp::getCStd(const std::string &std)
+simplecpp::cstd_t simplecpp::getCStd(const std::string &std, cstd_t dflt)
```
danmar: we dont need the dflt parameters anymore right?

glankk: Yeah, it can be removed

simplecpp.cpp
```diff
@@ -727,10 +730,63 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
...
+
                 if (!ppTok || !ppTok->number)
                     continue;
 
-                const unsigned int line = std::atol(ppTok->str().c_str());
+                unsigned long line;
+                try {
+                    line = std::stoul(ppTok->str());
+                } catch (...) {
+                    line = std::numeric_limits<unsigned long>::max();
```
danmar: this value might not work well.. it means tokens might get line number 0 which cppcheck interprets as "no line".

I suggest that the value passed in the `lineDirective` call will be 1. After reporting the error message.. change it to 1?

danmar: or assign `maxvalue+1`?

glankk: Sounds good.

test.cpp
```diff
@@ -2847,6 +2857,221 @@ static void nullDirective3()
...
+                          toString(outputList));
+        }
+
+        {
+            const char code[] =
+                "#line 18446744073709551617\n" // 2^64 + 1
+                ";\n";
+            simplecpp::OutputList outputList;
+            makeTokenList(code, dui, &outputList);
+            ASSERT_EQUALS("file0,2,portability_line_directive,Line number out of range: 18446744073709551617. Line numbers above 2147483647 are conditionally supported in " + std_name + ".\n",
```
danmar, line 3069: In theory this test will behave differently on a platform where `unsigned long` is 32 bits. I.e. on WIN32.
I don't have a strong opinion.. but it's a bit unfortunate.

simplecpp.cpp
glankk: Actually it might be better that we just ignore the line number and use the current one when it's out of range, what do you think?

glankk: Or just discard the line directive entirely.

test.cpp
glankk: `std::stoul` which is used for parsing line numbers does not overflow but throws an out of range exception. So this will behave the same everywhere `unsigned long` is at least 32 bits.

glankk: I assume that's what you're referring to, if you mean the line number in the warning then it's just the token string.

danmar: I meant the "18446744073709551617" in the expected string. The assertion will fail because the output will say ~ 4 billion if you run the test on a WIN32 platform..

simplecpp.cpp
danmar: I would be fine with both those options.

test.cpp
glankk: The output is just the token string, it's unaffected by the platform.

simplecpp.cpp
glankk: I've changed it to just use the maximum possible line number, I think that's alright. It's guaranteed to always be representable, and it's also the most lenient option in terms of conditional support.

ludviggunne (approved):

PR #689, opened 2026-07-23
Fix #688: #ifdef and #ifndef directives missing from ifCond

ludviggunne:


simplecpp.cpp
```diff
@@ -3740,11 +3740,23 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
                     conditionIsTrue = false;
                 }
                 else if (rawtok->str() == IFDEF) {
-                    conditionIsTrue = (macros.find(rawtok->next->str()) != macros.end() || (hasInclude && rawtok->next->str() == HAS_INCLUDE));
-                    maybeUsedMacros[rawtok->next->str()].emplace_back(rawtok->next->location);
+                    const std::string &name = rawtok->next->str();
+                    conditionIsTrue = (macros.find(name) != macros.end() || (hasInclude && name == HAS_INCLUDE));
+                    maybeUsedMacros[name].emplace_back(rawtok->next->location);
+                    if (ifCond) {
+                        std::string E = "defined(" + name + ")";
```
ludviggunne: An alternative to this is to update the `IfCond` struct to indicate what type of directive was evaluated.

danmar (approved):

simplecpp.cpp
danmar: I don't have strong opinions. Off-the-top-of-my-head I am not sure exactly how it's used.. it's some misra checkers but don't know exactly how..

simplecpp.cpp
```diff
@@ -3740,11 +3740,23 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
                     conditionIsTrue = false;
                 }
                 else if (rawtok->str() == IFDEF) {
-                    conditionIsTrue = (macros.find(rawtok->next->str()) != macros.end() || (hasInclude && rawtok->next->str() == HAS_INCLUDE));
-                    maybeUsedMacros[rawtok->next->str()].emplace_back(rawtok->next->location);
+                    const std::string &name = rawtok->next->str();
+                    conditionIsTrue = (macros.find(name) != macros.end() || (hasInclude && name == HAS_INCLUDE));
+                    maybeUsedMacros[name].emplace_back(rawtok->next->location);
+                    if (ifCond) {
+                        const std::string E = "defined(" + name + ")";
```
danmar: I think "E" should be the preprocessed condition.. i.e. it might be better if it just says "0" or "1" ?

danmar (approved):

PR #691, opened 2026-07-23
Fix #688: Support `#elifdef` and `#elifndef`

ludviggunne:


simplecpp.cpp
```diff
@@ -3721,7 +3725,8 @@ void simplecpp::preprocess(simplecpp::TokenList &output, const simplecpp::TokenL
                     rawtok = filedata->tokens.cfront();
                     continue;
                 }
-            } else if (rawtok->str() == IF || rawtok->str() == IFDEF || rawtok->str() == IFNDEF || rawtok->str() == ELIF) {
+            } else if (rawtok->str() == IF || rawtok->str() == IFDEF || rawtok->str() == IFNDEF || rawtok->str() == ELIF ||
+                       ((getCStd(dui.std) >= C23 || getCppStd(dui.std) >= CPP23) && (rawtok->str() == ELIFDEF || rawtok->str() == ELIFNDEF))) {
```
danmar: I feel that we don't need to check the standard. If we see `#elifdef` we can use the same handling used in C23/C++23.. Let's check the standard if some different behaviors are wanted.

test.cpp
```diff
@@ -2370,6 +2370,76 @@ static void elif()
...
+                            "#elifdef X\n"
+                            "2\n"
+                            "#else\n"
+                            "3\n"
+                            "#endif";
+        ASSERT_EQUALS("\n1", preprocess(code));
+    }
+    {
+        const char code[] = "#define X\n"
+                            "#if 0\n"
```
danmar, line 2387: test if elifdef is handled properly after a true scope.. i.e. if this line is `#ifdef X` instead.

ludviggunne: That's the test above

danmar (approved):

PR #692, opened 2026-07-24
Improve no_eof_newline warning

glankk:
This is an addendum to #672. Use `DUI::std` to decide if a warning should be produced, clarify error message, add more tests.

### DIFF
--- a/simplecpp.cpp
+++ b/simplecpp.cpp
@@ -1022,11 +1022,11 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
             location.adjust(currentToken);
     }
 
-    if (!trailing_nl && outputList) {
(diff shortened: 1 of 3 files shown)


simplecpp.cpp
```diff
@@ -1022,11 +1022,11 @@ void simplecpp::TokenList::readfile(Stream &stream, const std::string &filename,
             location.adjust(currentToken);
     }
 
-    if (!trailing_nl && outputList) {
+    if ((cstd != CUnknown || (cstd == CUnknown && cppstd == CPPUnknown)) && !trailing_nl && outputList) {
```
danmar: technically there is redundant code, it can be simplified to:
```suggestion
    if ((cstd != CUnknown || cppstd == CPPUnknown) && !trailing_nl && outputList) {
```

danmar (approved):

danmar (approved):

PR #694, opened 2026-08-05
Fix #653: defined() not recognizing ## token paste in its operand (GNU extension)

ludviggunne:


simplecpp.cpp
```diff
@@ -2249,33 +2249,38 @@ namespace simplecpp {
...
+                        macroName += temp.cback() ? temp.cback()->str() : "";
+                    else
+                        macroName += argTok->str();
+                    argTok = argTok->next;
+                    if (!sameline(tok, argTok))
+                        break;
+                    if (argTok && argTok->str() == "#") {
+                        if (!argTok->next || argTok->next->str() != "#")
+                            throw Error(argTok->location, "Expected '#'");
+                        argTok = argTok->next->next;
```
danmar, line 2271: is there any possibility that `argTok->next->next` is null here? then `while (argTok->name)` would cause a null pointer dereference.

simplecpp.cpp
```diff
@@ -2249,33 +2249,38 @@ namespace simplecpp {
...
-                        lastToken = defToken->next->next->next;
+                while (argTok->name) {
+                    if (expandArg(temp, argTok, parametertokens))
+                        macroName += temp.cback() ? temp.cback()->str() : "";
+                    else
+                        macroName += argTok->str();
+                    argTok = argTok->next;
+                    if (!sameline(tok, argTok))
+                        break;
+                    if (argTok && argTok->str() == "#") {
```
danmar: as far as I remember, if argTok is null then sameline returns false. so argTok can't be null here.

simplecpp.cpp
```diff
@@ -2249,33 +2249,40 @@ namespace simplecpp {
...
-                        TokenList temp(files);
-                        if (expandArg(temp, defToken, parametertokens))
-                            macroName = temp.cback()->str();
-                        if (expandArg(temp, defToken->next->next->next, parametertokens))
-                            macroName += temp.cback() ? temp.cback()->str() : "";
-                        else
-                            macroName += defToken->next->next->next->str();
-                        lastToken = defToken->next->next->next;
+                while (argTok->name) {
+                    if (expandArg(temp, argTok, parametertokens))
```
danmar, line 2261: I am not sure that we should expand macros unless they are operands. will this work:
```
#define x
#if defined(x)
```
the `macroName` will be empty string here now? and we will think that the condition is false?

ludviggunne: It works, it takes a different execution path.

simplecpp.cpp
```diff
@@ -2249,33 +2249,40 @@ namespace simplecpp {
...
-                if (sameline(tok, tok4) && tok2->op == '(' && tok3->name && tok4->op == ')') {
-                    defToken = tok3;
-                    lastToken = tok4;
-                } else if (sameline(tok,tok2) && tok2->name) {
-                    defToken = lastToken = tok2;
+                std::string macroName;
+                TokenList temp(files);
+                const Token *argTok = tok->next;
+                bool expectParen = false;
+                if (argTok->op == '(') {
```
danmar, line 2256: we used to prevent null pointer dereference when `tok->next` is null

danmar: it seems that sameline is not used properly anymore.

PR #698, opened 2026-08-07
Add ending newlines to all test cases where applicable

glankk:
Adds more newlines to test cases. These aren't problematic at the moment because the output is ignored for these cases, but I think it's good to have a consistent style with regards to ending test cases code with newlines.

### DIFF
--- a/test.cpp
+++ b/test.cpp
@@ -267,10 +267,10 @@ static void backslash()
 
 static void builtin()
 {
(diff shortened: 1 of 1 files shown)


danmar (commented):
hmm.. an alternative could be that `preprocess` automatically appends a newline by default.
We can have a separate `preprocessNoNewline` that does not append the newline.
what do you think?

